PR #1, opened 2018-08-15
add local development helpers

phoracek:
Ready to be reviewed

### DIFF
--- a/cluster/cli.sh
+++ b/cluster/cli.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+#
+# Copyright 2018 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -e
+
+source ./cluster/gocli.sh
+
+$gocli_interactive ssh node01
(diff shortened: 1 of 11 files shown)


cluster/up.sh
```diff
@@ -0,0 +1,38 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -e
+
+source ./cluster/gocli.sh
+
+$gocli run --random-ports --nodes 2 --background kubevirtci/k8s-1.10.3
```
yuvalif: one node or 2 nodes?

cluster/cli.sh
```diff
@@ -0,0 +1,22 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -e
+
+source ./cluster/gocli.sh
+
+$gocli_interactive ssh node01
```
yuvalif: is node01 hardcoded somewhere?

cluster/up.sh
phoracek: Bad copy-paste, fixing

cluster/cli.sh
phoracek: Now when number of nodes is fixed, this is the only one we got

yuvalif: why not copy the standard CI env from kubevirt?

phoracek:
How would that look like?

cluster/cli.sh
phoracek: How would that look like?

phoracek: I tried to reference kubevirt cluster handling, but it will not be that easy. We can use it's `up` and `down`, but the rest is kubevirt specific. We cannot easily combine them, at least I don't see how.

@yuvalif is it ok if we keep this simple environment for now? I'd like to unblock work on the actual plugin, we can improve this in parallel.

@SchSeba PR for what? For the multi-node and VXLAN? We could. It can even be just a simple VLAN, but I guess it does not matter which one we use.

yuvalif: @phoracek we can track that as an enhancement issue.

phoracek:
Opened an issue #3 for advanced local cluster, merging and closing this one.

PR #4, opened 2018-08-16
add basic OVS CNI plugin implementation

### LINKED ISSUE #2: Basic OVS CNI plugin implementation
With this initial implementation, ovs-cni would just connect container to an OVS bridge available on host and optionally assign connection port to a VLAN. Support for IPAM and other advanced features will be added later and described in separate issues. Pure L2 plugins sounds limiting, but it should be enough for KubeVirt VMs use-case.

This basic support should reflect bridge plugin from containernetworking/plugins. Code must have unit tests coverage.

# Open vSwitch CNI Plugin

## Overview

With ovs plugin, containers (on the same host) are plugged into an Open vSwitch
bridge (virtual switch) that resides in the host network namespace. It's host
adminitrator's responsibility to create such bridge and optinally connect it to
broader network, be it using L2 directly, NAT or an overlay. The containers
receive one end of the veth pair with the other end connected to the bridge.

## Example configuration

```json
{
    "name": "mynet",
    "type": "ovs",
    "bridge": "mynet0",
    "vlan": 100
}
```

## Network configuration reference

* `name` (string, required): the name of the network.
* `type` (string, required): "ovs".
* `bridge` (string, optional): name of the bridge to use/create. Defaults to
   "ovsbr0".
* `vlan` (integer, optional): VLAN ID of attached port. Opened for all if not
   specified.

phoracek:
Resolves #2.

Just connect container to an OVS bridge and give its port specified VLAN ID.

* [x] Plugin implemented
* [x] Plugin documentation
* [x] Example for manual testing
* [x] Code comments
* [x] Logging
* [x] Tests
* [x] Deployment on local cluster
* [x] Usage example on the local cluster


phoracek:
@dankenigsberg @yuvalif @SchSeba could you please review this? Thanks.

docs/cni-plugin.md
```diff
@@ -0,0 +1,91 @@
...
+    "vlan": 100
+}
+```
+
+## Network Configuration Reference
+
+* `name` (string, required): the name of the network.
+* `type` (string, required): "ovs".
+* `bridge` (string, required): name of the bridge to use.
+* `vlan` (integer, optional): VLAN ID of attached port. Opened for all if not
```
SchSeba: s/Opened for all/Trunk port/

docs/cni-plugin.md
```diff
@@ -0,0 +1,91 @@
...
+## Manual Testing
+
+```shell
+# Build the binary
+make build-plugin
+
+# Create a new namespace
+ip netns add ns1
+
+# Create OVS bridge on the host
```
SchSeba: add here how to install ovs or maybe just a note that ovs needs to be installed and running

docs/cni-plugin.md
```diff
@@ -0,0 +1,91 @@
...
+# Delete OVS bridge
+ovs-vsctl del-br br1
+
+# Delete the namespace
+ip netns del ns1
+```
+
+## Go Tests
+
+This plugin also have Go test coverage. To run tests, Open vSwitch must be
```
SchSeba: Add this to the manual section maybe :)

docs/local-cluster-example.md
```diff
@@ -0,0 +1,82 @@
+# Usage Example on Local Cluster
```
SchSeba: We can use the provider I already created on the kubevirtci project? (KUBEVIRT_PROVIDER=k8s-multus-1.11.1)

Then the cluster-sync only needs to deploy the ovs plugin daemonset.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,278 @@
...
+
+func loadNetConf(bytes []byte) (*netConf, error) {
+	netconf := &netConf{}
+	if err := json.Unmarshal(bytes, netconf); err != nil {
+		return nil, fmt.Errorf("failed to load netconf: %v", err)
+	}
+
+	if netconf.BrName == "" {
+		return nil, fmt.Errorf("\"bridge\" is a required argument")
+	}
```
SchSeba: Also can we make a vlan validation?

phoracek: It is later validated by ovs-vsctl `ovs-vsctl: constraint violation: 5555 is not in the valid range 0 to 4095 (inclusive)`, I think we should validate only necessary minimum, but I'm open to discuss it.

docs/local-cluster-example.md
phoracek: Donw

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,278 @@
...
+	types.NetConf
+	BrName  string `json:"bridge"`
+	VlanTag *uint  `json:"vlan"`
+}
+
+func init() {
+	// this ensures that main runs only on main thread (thread group leader).
+	// since namespace ops (unshare, setns) are done for a single thread, we
+	// must ensure that the goroutine does not jump from OS thread to thread
+	runtime.LockOSThread()
```
yuvalif: where do you use network actions inside goroutines?

Issue #6, opened 2018-08-24
Not all supported platforms offer Go 1.10 or higher


phoracek:
In the current state, with using network namespaces from Go, we enforce Go 1.10, since this code cannot be compiled on Go 1.9 or lower [1]. However, at this moment we need to work with Go 1.9 to support all platforms.

We should create a temporary workaround to work in container network namespace using `nsenter` or `ip netns`.

This code should be reverted and issue closed as soon as Go 1.9 platforms support is dropped.

[1] https://www.weave.works/blog/linux-namespaces-and-go-don-t-mix

SchSeba:
Can we leave the current implementation on a new branch?

phoracek:
We could, but not sure if it's worth it. Almost finished the new
implementation and it's not that big of a change, plus we have got revert.

SchSeba:
OK works for me

phoracek:
WIP PR pushed https://github.com/kubevirt/ovs-cni/pull/7

yuvalif:
which platforms don't support go.1.10? seems like mac/win/linux are supported: https://golang.org/dl/

phoracek:
This repository master will require Go 1.10, 1.9 will be handled on a separate private branch to not block further development.

PR #7, opened 2018-08-27
use nsenter to access container namespace

phoracek:
With Go 1.9 or older, namespaces does not work properly. With this patch we execute a separate process (nsenter command line) every time we access container namespace.

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -13,10 +13,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Go version 1.10 or greater is required. Before that, switching namespaces in
-// long running processes in go did not work in a reliable way.
-// +build go1.10
-
 package plugin
 
 import (
@@ -25,12 +21,12 @@ import (
 	"log"
 	"os/exec"
 	"runtime"
+	"strings"
 
 	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types"
 	"github.com/containernetworking/cni/pkg/types/current"
 	"github.com/containernetworking/plugins/pkg/ip"
-	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/vishvananda/netlink"
 )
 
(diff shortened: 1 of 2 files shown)


phoracek:
@yuvalif @SchSeba hi, could you please take a look?

pkg/plugin/plugin.go
```diff
@@ -266,13 +266,10 @@ func CmdDel(args *skel.CmdArgs) error {
 
 	// Delete can be called multiple times, so don't return an error if the
 	// device is already removed.
-	err = ns.WithNetNSPath(args.Netns, func(ns.NetNS) error {
-		err = ip.DelLinkByName(args.IfName)
-		if err != nil && err == ip.ErrLinkNotFound {
-			return nil
-		}
+	_, err = withNetNS(args.Netns, "ip", "link", "del", "dev", args.IfName)
+	if err != nil && strings.Contains(err.Error(), "Cannot find device") {
```
yuvalif: the logic is inverse, if you have an error which is not "cannot find device" then return an error

pkg/plugin/plugin.go
```diff
@@ -177,11 +182,12 @@ func CmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
-	// Refetch the bridge since its MAC address may change when the first
-	// veth is added.
-	if err = refetchIface(brIface); err != nil {
-		return err
+	// Refetch the bridge MAC since it may change when the first veth is added.
```
yuvalif: how is this change related to the ns fix?

yuvalif:
not sure regarding issue #6 
which supported platforms don't support golang10 ?

pkg/plugin/plugin.go
```diff
@@ -86,32 +82,56 @@ func setupBridge(brName string) (*current.Interface, error) {
...
 
-	// Refetch the hostIface since its MAC address may change during network namespace move.
-	if err = refetchIface(hostIface); err != nil {
+	contIface.Name = contIfaceName
+	contIface.Sandbox = contNetnsPath
+	contIfaceMac, err := getIfaceMac(contNetnsPath, contIfaceName)
+	if err != nil {
 		return nil, nil, err
 	}
+	contIface.Mac = contIfaceMac
```
SchSeba: small nit don't need a new variable here just use `contIface.Mac`

pkg/plugin/plugin.go
SchSeba: Need to be `if err != nil && !(strings.Contains(err.Error(), "Cannot find device"))`

phoracek:
This repository master will require Go 1.10, 1.9 will be handled on a separate private branch to not block further development.

PR #13, opened 2018-08-28
update documentation

### LINKED ISSUE #11: Document deployment and usage on arbitrary cluster
Currently, we document only usage on local kubevirtci cluster. We should add a description how to install multus and ovs-cni on a cluster and how to use it. This depends on #10

phoracek:
Resolves #11


README.md
```diff
@@ -1,5 +1,48 @@
 # Open vSwitch CNI plugin
 
+This plugin allows user to define Kubernetes networks on top of Open vSwitch bridges available on nodes. IPAM is currently not supported. There is no scheduling involved, desired bridges must be precreated on all nodes. Also, ovs-cni does not configure bridges, it's up to user to connect them to L2, L3 or an overlay network. Finally please note that Open vSwitch must be installed and running on the host.
```
SchSeba: Maybe add here that also the network attachment CRD needs to be applied and multus cni plugin configured on every node in the cluster.

README.md
```diff
@@ -1,5 +1,48 @@
...
+
+## Usage
+
+First define network definition. This object specifies to which Open vSwitch bridge should be pod attached and what VLAN ID should be set on the port. For more information, check [plugin documentation](docs/cni-plugin.md).
```
yuvalif: "First define network definition"->"First create network attachment definition"

yuvalif: "should be pod"->"should the pod be"

README.md
```diff
@@ -1,5 +1,48 @@
...
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: ovs-conf
+spec:
+  config: '{
+      "cniVersion": "0.3.1",
+      "type": "ovs",
+      "bridge": "br1",
+      "vlan": 100
```
yuvalif: is vlan optional? can we define a port without specifying a vlan?

README.md
```diff
@@ -1,5 +1,48 @@
...
+  name: samplepod
+  annotations:
+    k8s.v1.cni.cncf.io/networks: ovs-conf
+spec:
+  containers:
+  - name: samplepod
+    command: ["/bin/bash", "-c", "sleep 99999"]
+    image: alpine
+EOF
+```
```
yuvalif: can you please add "ip link.." command to execute on the pod to verify that the interface was created

README.md
yuvalif: could you also add sample command to create the ovs bridge?

yuvalif: at the end of the readme (appendix?) may be interesting to add example of manual configuration of host labels to allow for host selection by the pod

README.md
phoracek: It's described in "For more information, check [plugin documentation](docs/cni-plugin.md)." Should I state it also here? I don't want to get into more and more details on from README.md as we add other paramters.

README.md
phoracek: It is already in docs/local-cluster-example.md, I will add it to detailed usage example for arbitrary cluster too. I'd prefer the from README.md to be as clean as possible.

phoracek: I will create another documentation file with this under docs/.

phoracek:
@SchSeba @yuvalif I reworked the documentation. Could you please take a look? Thanks

README.md
yuvalif: "local cluster" is for developers. the README should be for users.
if you want to keep it clean, maybe refer to "getting started" page?

yuvalif: covered in: deployment-on-arbitrary-cluster.md

PR #16, opened 2018-09-05
Add STDCI

gbenhaim:
Signed-off-by: gbenhaim <galbh2@gmail.com>

### DIFF
--- a/automation/check-merged.environment.yaml
+++ b/automation/check-merged.environment.yaml
@@ -1,0 +1,12 @@
+---
+- name: QUAY_REGISTRY_PASSWORD
+  valueFrom:
+    secretKeyRef:
+      name: kubevirt_quay
+      key: stdci_bot_password
+
+- name: QUAY_REGISTRY_USERNAME
+  valueFrom:
+    secretKeyRef:
+      name: kubevirt_quay
+      key: stdci_bot_username
--- a/automation/check-patch.packages
+++ b/automation/check-patch.packages
@@ -1,0 +1,2 @@
+docker
+make
--- a/automation/check-merged.mounts
+++ b/automation/check-merged.mounts
@@ -1,0 +1,1 @@
+check-patch.mounts
--- a/automation/check-merged.packages
+++ b/automation/check-merged.packages
@@ -1,0 +1,1 @@
+check-patch.packages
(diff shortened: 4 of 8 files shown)


automation/check-patch.packages
```diff
@@ -0,0 +1,2 @@
+docker
+golang-1.9.4
```
SchSeba: we need to use golang > 1.9

gbenhaim:
@SchSeba I removed the go installation.

SchSeba:
@gbenhaim can we configure it just like kubevirt?
```
git
rsync
docker
make
```

phoracek:
ci test please

automation/check-merged.environment.yaml
```diff
@@ -0,0 +1,5 @@
+- name: REGISTRY_PASSWORD
+  valueFrom:
+    secretKeyRef:
+      name: quay
+      key: kubevirt_bot
```
phoracek: Once there is a bot created for CI, would you please add just `docker login` to check-merged.sh? We will add the rest then.

gbenhaim: Sure. Can we do it in a different PR?

gbenhaim:
@SchSeba I've updated to packages list.

automation/check-patch.packages
```diff
@@ -0,0 +1,4 @@
+git
```
phoracek: Do we need all of these now? Please drop unused dependencies.

automation/check-merged.environment.yaml
phoracek: Fair enough

automation/check-patch.packages
gbenhaim: It was requested by @SchSeba

SchSeba: @phoracek I think the only one we can drop is git no?

phoracek: Also I don't see rsync and make used.

SchSeba: Ok lets try.
This was copy from the kubevirt repo.

gbenhaim:
@phoracek I've managed to squeeze the login to quay to this PR.
I think that we can merge it.

phoracek:
@gbenhaim that's perfect! Thanks a lot.

PR #17, opened 2018-09-06
Add host-network to daemonset.

SchSeba:
Add host-network to daemonset to allow it to start even if the primary network is not ready.

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/ovs-cni/17)

### DIFF
--- a/examples/ovs-cni.yml
+++ b/examples/ovs-cni.yml
@@ -14,6 +14,7 @@ spec:
         tier: node
         app: ovs-cni-plugin
     spec:
+      hostNetwork: true
       nodeSelector:
         beta.kubernetes.io/arch: amd64
       tolerations:
--- a/examples/ovs-cni.yml
+++ b/examples/ovs-cni.yml
@@ -31,6 +31,8 @@ spec:
           limits:
             cpu: "100m"
             memory: "50Mi"
+        securityContext:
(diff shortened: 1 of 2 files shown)


phoracek:
What happens without host network if the primary network is not ready?

examples/ovs-cni.yml
```diff
@@ -14,6 +14,7 @@ spec:
         tier: node
         app: ovs-cni-plugin
     spec:
+      hostNetwork: true
```
phoracek: We should add it to cluster/examples/ too

SchSeba:
> What happens without host network if the primary network is not ready?

The pod stuck in pending just like the coreDNS.

examples/ovs-cni.yml
SchSeba: Done

SchSeba:
This will also fix #18

examples/ovs-cni.yml
```diff
@@ -30,6 +31,8 @@ spec:
           limits:
             cpu: "100m"
             memory: "50Mi"
+        securityContext:
```
phoracek: Why do we need to raise privileges?

phoracek:
Thanks for testing it on OpenShift

PR #19, opened 2018-09-12
explicitly set container interface mac to local unicast

phoracek:
With prefix 02:00:00:, we get local unicast MAC address from non-reserved range.

Addresses https://github.com/kubevirt/kubevirt/issues/1494

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -23,6 +23,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"math/rand"
+	"net"
 	"os/exec"
 	"runtime"
 
@@ -33,6 +35,8 @@ import (
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/vishvananda/netlink"
 )
+
+const macSetupRetries = 2
 
 type netConf struct {
 	types.NetConf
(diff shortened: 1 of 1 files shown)


phoracek:
Passes plugin test and assigns MAC with 02:00:00 prefix to container side of veth pair.

@SchSeba @yuvalif could you please review?

pkg/plugin/plugin.go
```diff
@@ -98,8 +110,20 @@ func setupVeth(contNetns ns.NetNS, contIfaceName string) (*current.Interface, *c
...
 			return err
 		}
+
+		containerLink, err := netlink.LinkByName(containerVeth.Name)
+		if err != nil {
+			return fmt.Errorf("failed to lookup %q: %v", containerVeth.Name, err)
+		}
+
+		containerMac := generateRandomMac()
+		err = netlink.LinkSetHardwareAddr(containerLink, containerMac)
```
yuvalif: will this prevent duplicates?

phoracek: It won't, but there are not many interfaces in the pod, do you think there is real risk of collisions?

phoracek: But with only 3 octets used, I guess it could happen. Fixing

phoracek:
@yuvalif added dirty fix for the collision, I feel bad (just a little).

phoracek:
@yuvalif sorry, I'm just bad at maths, there are pretty low chances of a collision.

pkg/plugin/plugin.go
```diff
@@ -98,8 +110,25 @@ func setupVeth(contNetns ns.NetNS, contIfaceName string) (*current.Interface, *c
...
 			return err
 		}
+
+		containerLink, err := netlink.LinkByName(containerVeth.Name)
+		if err != nil {
+			return fmt.Errorf("failed to lookup %q: %v", containerVeth.Name, err)
+		}
+
+		// In case the MAC address is already assigned to another interface, retry
+		macSetupRetries := 2
```
SchSeba: Can you configure this as a const?

phoracek:
@SchSeba done

PR #20, opened 2018-09-14
set bridge port UP after attaching pod

phoracek:
We need to set bridge ports explicitly UP in order to be able to ping between two pods.


yuvalif:
i see it is already merged, but please add a test to cover that,

PR #21, opened 2018-09-14
installation manifests

phoracek:
Keep templates of installation manifests under the repository.

### DIFF
--- a/docs/deployment-on-arbitrary-cluster.md
+++ b/docs/deployment-on-arbitrary-cluster.md
@@ -11,9 +11,11 @@ local cluster](deployment-on-local-cluster.md) guide.
 
 ## Open vSwitch
 
-First of all, Open vSwitch must be installed and running on all nodes (unless
-you do [scheduling](scheduling.md)). OVS is available in repositories of all
-major distributions.
+### Kubernetes
+
+Open vSwitch must be installed and running on all nodes (unless you do
+[scheduling](scheduling.md)). OVS is available in repositories of all major
+distributions.
 
 On CentOS:
 
@@ -39,7 +41,18 @@ ovs-vsctl list-br
 ovs-vsctl del-br test-br
 ```
 
+### OpenShift
+
+With OpenShift you can either decide to install Open vSwitch from your
+repositories as described in previous section or using a DaemonSet.
+
+```shell
+kubectl apply -f https://raw.githubusercontent.com/kubevirt/ovs-cni/master/examples/openshift-ovs-vsctl.yml
+```
+
 ## Multus
+
+### Kubernetes
 
 Installation of Multus is currently a little tricky. If you have not installed
 any network plugin (Calico, Flannel,...) on your cluster, you can follow [quick
@@ -48,6 +61,15 @@ However, if you already have a network plugin running, you need to do some
 extra work. I would recommend you to study the quickstart guide and modify it
 for your own case. Multus should soon include better documentation of the
 installation process.
+
+### OpenShift
+
+Installation of Multus on OpenShift should be easier. Please note that this is
+not officially supported by OpenShift, it is only a dev preview with no warranties.
+
+```shell
+kubectl apply -f https://raw.githubusercontent.com/kubevirt/ovs-cni/master/examples/multus.yml
+```
 
 ## Open vSwitch CNI plugin
 
--- a/examples/openshift-ovs-vsctl.yml
+++ b/examples/openshift-ovs-vsctl.yml
@@ -1,7 +1,7 @@
 apiVersion: extensions/v1beta1
 kind: DaemonSet
 metadata:
-  name: ovs-cni-plugin-amd64
+  name: ovs-vsctl-amd64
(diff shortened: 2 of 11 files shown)


phoracek:
Verified, installed and tested ovs-cni using manifests under examples/.

examples/multus.yml
```diff
@@ -0,0 +1,134 @@
+---
```
yuvalif: why do we need this file under our examples? how is it different than the one taken from official multus site?

docs/deployment-on-arbitrary-cluster.md
```diff
@@ -49,6 +62,14 @@ extra work. I would recommend you to study the quickstart guide and modify it
 for your own case. Multus should soon include better documentation of the
 installation process.
 
+### OpenShift
```
yuvalif: which openshift version suport multus? don't think this is officially out yet. please check

docs/deployment-on-arbitrary-cluster.md
```diff
@@ -39,8 +41,19 @@ ovs-vsctl list-br
...
+
+```shell
+kubectl apply -f https://raw.githubusercontent.com/kubevirt/ovs-cni/master/examples/openshift-ovs-vsctl.yml
+```
+
 ## Multus
 
+### Kubernetes
+
 Installation of Multus is currently a little tricky. If you have not installed
```
yuvalif: what do you mean by tricky?
is there anything else to do than to reorder conf files under ```/etc/cni/net.d``` ?

examples/networkattachmentdefinition.yml
```diff
@@ -0,0 +1,11 @@
+apiVersion: "k8s.cni.cncf.io/v1"
```
yuvalif: please rename this file to ovs-net-vlan100.yaml and also the name to "ovs-net-vlan100".
there could be multiple network attachment definitions, even of the same plugin.

examples/pod.yml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: samplepod-1
+  annotations:
+    k8s.v1.cni.cncf.io/networks: ovs-net-vlan
```
yuvalif: please rename the network

examples/openshift-ovs-vsctl.yml
```diff
@@ -0,0 +1,50 @@
+apiVersion: extensions/v1beta1
```
yuvalif: why do we put this under "examples", this is part of the "installation" of the product. where do we place it for kubevirt?

examples/ovs-cni.yml
```diff
@@ -1,18 +1,17 @@
----
 apiVersion: extensions/v1beta1
```
yuvalif: is it an example or part of installation?

examples/multus.yml
phoracek: For OpenShift, changed its name.

docs/deployment-on-arbitrary-cluster.md
phoracek: It is only dev preview, not officially supported. I added a comment on that.

docs/deployment-on-arbitrary-cluster.md
phoracek: You need to add more configuration, like credentials for multus on Kubernetes API. And the delegate in net.d/00-multus.conf depends on primary network plugin you use.

examples/networkattachmentdefinition.yml
phoracek: Done

examples/pod.yml
phoracek: Done

examples/openshift-ovs-vsctl.yml
phoracek: In KubeVirt, generated examples are not kept under the repo IIUIC, you need to call make. Not 100% sure.

We keep it under examples/ since user might want to change image source/version, namespaces etc.

examples/ovs-cni.yml
phoracek: It can be used for installation if you want to use kube-system and :latest image.

examples/openshift-ovs-vsctl.yml
yuvalif: in kubevirt, it seems like the "installation" manifest is generated based on a template:
https://github.com/kubevirt/kubevirt/blob/8112d64eda725594210ca7f0255a1343b6d4904e/manifests/release/kubevirt.yaml.in
and generated under the _out directory

phoracek: @yuvalif would you prefer to keep them in the repo like this PR proposes or in temporary _out then?

yuvalif: I'm not even sure how the template is parametrized in kubevirt...
So, I think "examples" are good. maybe add a readme which explain what in the yaml could be changed? or maybe comments in the yaml?

PR #24, opened 2018-09-17
explicitly set container interface mac to local unicast

phoracek:
With prefix 02:00:00:, we get local unicast MAC address from non-reserved range.

Addresses kubevirt/kubevirt#1494


SchSeba:
The change is just like the master

PR #26, opened 2018-09-18
Revert "explicitly set container interface mac to local unicast"

phoracek:
This reverts commit e671b51a25017a0a9ccfb54e62fc19b263bd9e5b.
    
The MAC problem does not affest Pods, only VMs when used with KubeVirt.
This problems happens in 2 % times. It will be added in future relase.


phoracek:
There were no conflicts and tests are still passing.

PR #30, opened 2018-09-22
Added docker build and push to stdci

SchSeba:


SchSeba:
@phoracek Can you take a look I cant find the problem of the go build command.

phoracek:
@SchSeba I will take a look

phoracek:
@SchSeba found the problem. STD-CI runs Go 1.9, our code has constraint for 1.10.

I'm thinking of including 1.9 compatibility code in this repository until 1.10 is available. It would make our lives easier. What do you think?

PR #31, opened 2018-09-25
WIP stdci build binary

phoracek:


phoracek:
The problem of #30 was in old Go version. This is no needed.

PR #34, opened 2018-09-27
Add func test to stdci

SchSeba:
Allow the STDCI to run our functional tests.


automation/check-patch.sh
```diff
@@ -1,7 +1,21 @@
...
-    echo "TODO: Add functional tests"
+    echo "Start ovs process"
+    /usr/share/openvswitch/scripts/ovs-ctl --system-id=random start
+
+    cd ..
+    mkdir -p go/src/kubevirt.io
+    mkdir -p go/pkg
+
+    ln -s $(pwd)/ovs-cni go/src/kubevirt.io/
+    ls -la go/src/kubevirt.io/ovs-cni
```
phoracek: you can drop this

automation/check-patch.sh
```diff
@@ -1,7 +1,21 @@
...
+
+    export GOPATH=$(pwd)/go
```
phoracek: move this right after line 9

automation/check-patch.sh
```diff
@@ -1,7 +1,21 @@
...
+    cd go/src/kubevirt.io/ovs-cni
+
+    echo "Run tests"
+    make && make test
```
phoracek: if we are already on root, you can do `make build test`

automation/check-patch.packages
```diff
@@ -2,3 +2,4 @@ docker
 make
 git
 go
+openvswitch
```
phoracek: please add a newline after this

automation/check-patch.sh
SchSeba: Done

PR #35, opened 2018-10-02
Added functional test to check for duplicate mac addresses.

SchSeba:
Testing the plugin I get the same mac address for multiple pods, added a test to check it.

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/ovs-cni/35)

### DIFF
--- a/pkg/plugin/plugin_test.go
+++ b/pkg/plugin/plugin_test.go
@@ -210,8 +210,59 @@ var _ = Describe("CNI Plugin", func() {
 				testAddDel(conf, false)
 			})
 		})
+		Context("random mac address on container interface", func() {
+			It("should create eth0 on two different namespace with different mac addresses", func() {
+				conf := fmt.Sprintf(`{
+				"cniVersion": "0.3.1",
+				"name": "mynet",
+				"type": "ovs",
+				"bridge": "%s",
+				"vlan": %d
+				}`, BRIDGE_NAME, VLAN_ID)
+
+				const IFNAME = "eth0"
+
+				By("Creating two temporary target namespace to simulate two containers")
+				targetNsOne, err := ns.NewNS()
+				Expect(err).NotTo(HaveOccurred())
+				defer targetNsOne.Close()
+				targetNsTwo, err := ns.NewNS()
+				Expect(err).NotTo(HaveOccurred())
+				defer targetNsTwo.Close()
+
+				By("Checking that both namespaces have different mac addresses on eth0")
+				resultOne := attach(targetNsOne, conf, IFNAME)
+				contOneIface := resultOne.Interfaces[2]
+
+				resultTwo := attach(targetNsTwo, conf, IFNAME)
+				contTwoIface := resultTwo.Interfaces[2]
+
+				Expect(contOneIface.Mac).NotTo(Equal(contTwoIface.Mac))
+			})
+		})
 	})
 })
+
+func attach(namespace ns.NetNS, conf, ifName string) *current.Result {
+	args := &skel.CmdArgs{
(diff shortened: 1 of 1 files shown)


pkg/plugin/plugin_test.go
```diff
@@ -210,6 +210,62 @@ var _ = Describe("CNI Plugin", func() {
 				testAddDel(conf, false)
 			})
 		})
+		Context("random mac address on container interface", func() {
+			It("should create eth0 on two different namespace with different mac addresses", func() {
```
phoracek: Chances that you will create two matching MAC addresses are low. Would you consider changing this test so it verifies that the MAC address is picked from our local unicast subnet? You can keep the duplicate validation there.

pkg/plugin/plugin_test.go
```diff
@@ -210,6 +210,62 @@ var _ = Describe("CNI Plugin", func() {
...
+				}`, BRIDGE_NAME, VLAN_ID)
+
+				const IFNAME = "eth0"
+
+				By("Creating two temporary target namespace to simulate two containers")
+				targetNsOne, err := ns.NewNS()
+				Expect(err).NotTo(HaveOccurred())
+				targetNsTwo, err := ns.NewNS()
+				Expect(err).NotTo(HaveOccurred())
+				defer targetNsOne.Close()
```
phoracek: this should happen right after line 227 (in case targetNsTwo fails to be created).

pkg/plugin/plugin_test.go
```diff
@@ -210,6 +210,61 @@ var _ = Describe("CNI Plugin", func() {
...
+					ContainerID: "dummy",
+					Netns:       targetNsOne.Path(),
+					IfName:      IFNAME,
+					StdinData:   []byte(conf),
+				}
+
+				var result *current.Result
+
+				By("Calling ADD command")
+				r, _, err := cmdAddWithArgs(args, func() error {
```
phoracek: Could you please keep code to call ADD and obtain Interface as inline function and then call it twice, each time for one of the namespaces? To improve readability.

SchSeba: Done

pkg/plugin/plugin_test.go
```diff
@@ -210,9 +210,60 @@ var _ = Describe("CNI Plugin", func() {
...
+				resultTwo := attach(targetNsTwo,conf,IFNAME)
+				contTwoIface := resultTwo.Interfaces[2]
+
+				Expect(contOneIface.Mac).NotTo(Equal(contTwoIface.Mac))
+			})
+		})
 	})
 })
 
+func attach(namespace ns.NetNS,conf, IFNAME string) (*current.Result) {
```
phoracek: Please run gofmt on the code. There are missing spaces after commas

SchSeba: Done

pkg/plugin/plugin_test.go
```diff
@@ -210,9 +210,60 @@ var _ = Describe("CNI Plugin", func() {
...
+				resultTwo := attach(targetNsTwo, conf, IFNAME)
+				contTwoIface := resultTwo.Interfaces[2]
+
+				Expect(contOneIface.Mac).NotTo(Equal(contTwoIface.Mac))
+			})
+		})
 	})
 })
 
+func attach(namespace ns.NetNS, conf, IFNAME string) *current.Result {
```
phoracek: Please use lower case for IFNAME.

SchSeba: Done.

Call the interface ifName

PR #43, opened 2018-10-08
Fix yaml files.

SchSeba:
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/ovs-cni/43)


SchSeba:
Tested on a openshift deployment

PR #47, opened 2018-10-21
Remove kubevirt from stdci deployment.

leongold:
We want to strictly test ovs-cni functionality.


leongold:
looks ready?

SchSeba:
Good to go

PR #50, opened 2018-11-20
[WIP] add device plugin

phoracek:
Add device plugin that will add available bridges as node resources.

When an OVS bridge is marked with:

```
ovs-vsctl set Bridge br1 external_ids:ovs-cni-resource=red
```

This resource will be exposed on node as `ovs-cni.network.kubevirt.io/red.

Multus webhook must be used to pull resource name from networka attachment definition to pod request.


phoracek:
Abandoned in favour of https://github.com/kubevirt/ovs-cni/pull/51

PR #56, opened 2018-12-06
[WIP] Fix unit and functional tests for the plugin

SchSeba:
* unit tests are going to run on travis.
* functional tests are going to run on STD-CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,12 @@ env:
 
 install: true
 script:
+  - apt-get install openvswitch -y
+  - systemctl start openvswitch
   - cd ../..
   - mv kubevirt ../kubevirt.io
   - cd ../kubevirt.io/ovs-cni
+  - make test
 
 deploy:
 - provider: script
--- a/tests/ovs_test.go
+++ b/tests/ovs_test.go
@@ -17,7 +17,6 @@ package tests_test
 
 import (
 	"flag"
-
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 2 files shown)


SchSeba:
@phoracek  Can you please take a look?

/hold cancel

Makefile
```diff
@@ -20,6 +20,9 @@ format:
 	go vet ./pkg/...
 
 test:
+	hack/tests.sh
+
+docker-test:
```
phoracek: it should be the other way around - docker-test runs hack/tests.sh

hack/tests.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
phoracek: please rename it to hack/test-dockerized.sh

hack/tests.sh
```diff
@@ -0,0 +1,42 @@
...
+# Reduce verbosity if an automated build
+BUILD_QUIET=
+if [ -n "$JOB_NAME" -o -n "$TRAVIS_BUILD_ID" ]; then
+    BUILD_QUIET="-q"
+fi
+
+# Build the build container
+(cd ${DOCKER_DIR} && docker build . ${BUILD_QUIET} -t ${BUILDER})
+
+docker run --rm -it --privileged --network host --cap-add ALL -v /lib/modules:/lib/modules -v `pwd`:/root/go/src/kubevirt.io/ovs-cni -w "/root/go/src/kubevirt.io/ovs-cni" kubevirt make docker-test
```
phoracek: what does 'kubevirt' stands here for?

hack/docker-builder/entrypoint.sh
```diff
@@ -4,4 +4,12 @@ set -o pipefail
 
 source /etc/profile.d/gimme.sh
 export GOPATH="/root/go"
+
+# launch OVS
+function quit {
```
phoracek: i don't like the fact that we use docker-**builder** to run tests, but i guess we can live with it if it saves us from another container

Makefile
SchSeba: Done

hack/docker-builder/entrypoint.sh
SchSeba: Yeah I don't know to create another container for that.

hack/tests.sh
SchSeba: Done

hack/tests.sh
SchSeba: this is the name from BUILDER variable change it to ${BUILDER} as is in the func-test-build file

pkg/plugin/plugin_test.go
```diff
@@ -16,17 +16,16 @@ package plugin
 
 import (
 	"fmt"
-	"net"
```
phoracek: this should remain here

phoracek:
Great work @SchSeba! Thanks. Once it passes STD-CI, we can merge

SchSeba:
@phoracek Done the openshift is broken but it is unrelated to this PR

PR #62, opened 2018-12-20
Assign interface to ovn port

AlonaKaplan:
The  ovn port name should be specified in the network pod annotation -
k8s.v1.cni.cncf.io/networks: '[
{
"name": "ovs-ovn-conf",
"mac": "0a:00:00:00:00:80"
"OvnPort" : "ovn_port_name"
}
]'

There is no need to specify the bridge name in the network attachment definotion since the
default ovn bridge is known  - "br-int".

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -40,8 +40,14 @@ const macSetupRetries = 2
 
 type netConf struct {
 	types.NetConf
-	BrName  string `json:"bridge"`
+	BrName  string `json:"bridge,omitempty"`
 	VlanTag *uint  `json:"vlan"`
+}
+
+type EnvArgs struct {
+	types.CommonArgs
+	MAC     types.UnmarshallableString `json:"mac,omitempty"`
+	OvnPort types.UnmarshallableString `json:"ovnPort,omitempty"`
 }
 
 func init() {
@@ -65,14 +71,22 @@ func assertOvsAvailable() error {
 	return nil
 }
 
+func getEnvArgs(envArgsString string) (*EnvArgs, error) {
+	if envArgsString != "" {
+		e := EnvArgs{}
+		err := types.LoadArgs(envArgsString, &e)
+		if err != nil {
+			return nil, err
+		}
+		return &e, nil
+	}
+	return nil, nil
+}
+
 func loadNetConf(bytes []byte) (*netConf, error) {
 	netconf := &netConf{}
 	if err := json.Unmarshal(bytes, netconf); err != nil {
 		return nil, fmt.Errorf("failed to load netconf: %v", err)
-	}
-
-	if netconf.BrName == "" {
-		return nil, fmt.Errorf("\"bridge\" is a required argument")
 	}
 
 	return netconf, nil
@@ -165,7 +179,17 @@ func assignMacToLink(link netlink.Link, mac net.HardwareAddr, name string) error
 	return nil
 }
 
-func attachIfaceToBridge(hostIfaceName string, contIfaceName string, brName string, vlanTag *uint, contNetnsPath string) error {
+func getBridgeName(bridgeName, ovnPort string) (string, error) {
+	if bridgeName != "" {
+		return bridgeName, nil
+	} else if bridgeName == "" && ovnPort != "" {
+		return "br-int", nil
+	}
+
+	return "", fmt.Errorf("failed to get bridge name")
+}
+
+func attachIfaceToBridge(hostIfaceName string, contIfaceName string, brName string, vlanTag *uint, contNetnsPath string, ovnPortName string) error {
 	// Set external IDs so we are able to find and remove the port from OVS
 	// database when CNI DEL is called.
 	command := []string{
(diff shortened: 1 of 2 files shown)


ovirt-infra:
Hello contributor, thanks for submitting a PR for this project! 

I am the bot who triggers "standard-CI" builds for this project.
As a security measure, I will not run automated tests on PRs that are not from white-listed contributors.

In order to allow automated tests to run, please ask one of the project maintainers to review the code and then do one of the following:

1. Type `ci test please` on this PR to trigger automated tests for it.
2. Type `ci add to whitelist` on this PR to trigger automated tests for it and also add you to the contributor white-list so that your future PRs will be tested automatically.
3. If you are planning to contribute to more than one project, maybe it's better to ask them to add you to the project organization, so you'll be able to run tests for all the organization's projects.

SchSeba:
ci add to whitelist

pkg/plugin/plugin.go
```diff
@@ -65,14 +71,22 @@ func assertOvsAvailable() error {
...
+			return nil, err
+		}
+		return &e, nil
 	}
+	return nil, nil
+}
 
-	if netconf.BrName == "" {
-		return nil, fmt.Errorf("\"bridge\" is a required argument")
+func loadNetConf(bytes []byte, bridgeIsRequired bool) (*netConf, error) {
```
phoracek: `bridgeIsRequired` should not be passed down to this level. Also, we cannot ignore failure without checking that it was indeed because of missing bridge argument.

We should set BrName as `omitempty` and check if it was provided or not in caller of `loadNetConf`

pkg/plugin/plugin.go
```diff
@@ -317,17 +334,34 @@ func removeOvsPort(bridge string, portName string) error {
...
 
-	netconf, err := loadNetConf(args.StdinData)
+	envArgs, err := getEnvArgs(args.Args)
 	if err != nil {
 		return err
 	}
 
-	if args.Netns == "" {
-		panic("This should never happen, if it does, it means caller does not pass container network namespace as a parameter and therefore OVS port cleanup will not work")
+	var ovnPort string
```
phoracek: This is confusing. We get ovsPort during removal just so we can eventually use "br-int" default. We should handle the defaulting in a separate function.

pkg/plugin/plugin_test.go
```diff
@@ -41,7 +41,8 @@ var _ = BeforeSuite(func() {
 })
 
 var _ = AfterSuite(func() {
-	exec.Command("ovs-vsctl", "del-br", "--if-exists", BRIDGE_NAME).Run()
+	exec.Command("ovs-vsctl", "del-br", BRIDGE_NAME).Run()
+	exec.Command("ovs-vsctl", "del-br", "br-int").Run()
```
phoracek: We should not do this. BRIDGE_NAME is a "test-bridge", it's unlikely we would break something for user. However, with this we might accidentally remove real "br-int" bridge.

If we move the defaulting into a separate function, we might test that function and don't actually remove anything.

SchSeba: I can't understand this comment can you please explain me in more details?

phoracek: Sure. So the problem could be: I run those tests directly on my laptop - with OVS running. This tests creates BRIDGE_NAME bridge and then removes it, that does not do any harm, since "test-bridge" is unlikely to be used for anything but tests. On the other hand, if I had OVN running on my machine, this test would use it and once it is done, it would remove it. What if "br-int" was
actually used there?

So what I suggest is. Move defaulting (if ovsPort is set and no bridge provided, set bridge to br-int) into a separate funtion. Then we can have a unit test for this function.

SchSeba:
@phoracek can you please take another look?

pkg/plugin/plugin.go
```diff
@@ -238,7 +254,16 @@ func CmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
-	brIface, err := setupBridge(netconf.BrName)
+	bridgeName, err := getBridgeName(netconf.BrName,ovnPort)
+	if err != nil {
+		return err
+	}
+
+	if bridgeName == "" {
```
phoracek: will this ever happen?

if `bridgeName == "" && ovnPort != ""`, then we return error in previous section. Tell me if I am missing something.

SchSeba: Good one

pkg/plugin/plugin.go
```diff
@@ -165,7 +179,17 @@ func assignMacToLink(link netlink.Link, mac net.HardwareAddr, name string) error
 	return nil
 }
 
-func attachIfaceToBridge(hostIfaceName string, contIfaceName string, brName string, vlanTag *uint, contNetnsPath string) error {
+func getBridgeName(bridgeName, ovnPort string) (string,error) {
```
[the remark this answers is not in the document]
SchSeba: Done

phoracek:
@SchSeba ready to merge?

SchSeba:
@phoracek Yes this is ready I am working on another PR to enable ovn provider for functional tests.

phoracek:
@AlonaKaplan @SchSeba great job :)

PR #65, opened 2019-02-10
fix empty resource reporting

phoracek:
With the current code, when there are no bridges on a node,
ovs-vsctl br-list returns "" which is later split to list of
bridges [""]. In order to prevent that, we need to check length
of returned string containing list of bridges.


phoracek:
@SchSeba could you please take a look? It's a minimal change.

phoracek:
ci test please

PR #68, opened 2019-03-05
Fixed `make build`

booxter:
The code imports from kubevirt.io/ovs-cni, but this path is not
available:

$ go get kubevirt.io/ovs-cni
package kubevirt.io/ovs-cni: unrecognized import path
"kubevirt.io/ovs-cni" (parse https://kubevirt.io/ovs-cni?go-get=1: no
go-import meta tags ())

But if I go get github.com/kubevirt/ovs-cni and then try to build it
then I get import failures.

This patch changes imports to use github.com/kubevirt/ovs-cni paths to
fix the build.

### DIFF
--- a/automation/check-patch.sh
+++ b/automation/check-patch.sh
@@ -20,10 +20,10 @@ main() {
     curl -sL https://raw.githubusercontent.com/travis-ci/gimme/master/gimme | HOME=/gimme bash >> /etc/profile.d/gimme.sh
     source /etc/profile.d/gimme.sh
 
-    mkdir -p $GOPATH/src/kubevirt.io
+    mkdir -p $GOPATH/src/github.com/kubevirt
     mkdir -p $GOPATH/pkg
-    ln -s $(pwd)/ovs-cni $GOPATH/src/kubevirt.io/
-    cd $GOPATH/src/kubevirt.io/ovs-cni
+    ln -s $(pwd)/ovs-cni $GOPATH/src/github.com/kubevirt/
+    cd $GOPATH/src/github.com/kubevirt/ovs-cni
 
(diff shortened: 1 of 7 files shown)


booxter:
/hold

More changes needed across build scripts to make it not break other stuff.

booxter:
@phoracek please check if I make sense here. I believe since ```kubevirt.io/ovs-cni``` redirection is not set, we should use full ```github.com/kubevirt/ovs-cni``` form for all imports.

SchSeba:
Hi @booxter can you explain where you encounter this problem?

* when you want to build the project move/clone it to ($GOPATH/kubevirt.io/ovs-cni)
* when you want to import it to another project using godep you can configure it like that
```toml
[[override]]
  name= "kubevirt.io/ovs-cni"
  source = "https://github.com/kubevirt/ovs-cni.git"
  version = "v0.1.0"
```

phoracek:
@booxter I am for this change, if @SchSeba is not strongly against. I think that kubevirt/kubevirt and kubevirt/ovs-cni are the only projects under kubevirt/ org using kubevirt.io as its import path. And it seems like kubevirt.io does not redirect to this project.

I don't think kubevirt.io in import path is worth additional difficulties Ihar described.

SchSeba:
@phoracek I don't have any strong feelings about it I am fine with both solutions

booxter:
@SchSeba it's my expectation that I should be able to build from checkout produced by ```go get ...```, which is not the case, whether I use ```kubevirt.io/ovs-cni``` or ```github.comkubevirt/ovs-cni```. (Before the PR merged.) I don't think one should be expected to move a directory around the disk to be able to build the project.

PR #69, opened 2019-03-05
Use `/usr/bin/env bash` instead of `/bin/bash` for build scripts

booxter:
Not every platform has bash available through /bin/bash (for example,
nixos doesn't put it there). But /usr/bin/env method is universal.


phoracek:
Thanks :)

PR #70, opened 2019-03-05
Don't copy ovs-vsctl to system directories

booxter:
On some platforms, for example, CoreOS, /usr/bin/ is not available for
writing, so an attempt to copy files there fails with 'read only file
system' error. Even if the file system would be mounted as read-write,
it's still a bad idea to touch it because this may potentially break
ovs-vsctl tool that may already be installed on the host.

Insted of writing and overriding files in /usr/bin/, copy them to
/opt/cni/bin that should be available for writing for all CNI plugin
pods.

### DIFF
--- a/examples/openshift-ovs-cni.yml
+++ b/examples/openshift-ovs-cni.yml
@@ -28,11 +28,11 @@ spec:
         - /bin/bash
         - -c
         - |
-          cp /usr/bin/ovs-vsctl /host/usr/bin/_ovs-vsctl
+          cp /usr/bin/ovs-vsctl /host/usr/local/bin/_ovs-vsctl
           echo '#!/usr/bin/bash
           _ovs-vsctl --db unix:///run/openvswitch/db.sock $@
-          ' > /host/usr/bin/ovs-vsctl
-          chmod +x /host/usr/bin/ovs-vsctl
+          ' > /host/usr/local/bin/ovs-vsctl
+          chmod +x /host/usr/local/bin/ovs-vsctl
         image: docker.io/openshift/origin-node:v3.10.0-rc.0
         imagePullPolicy: IfNotPresent
         resources:
@@ -45,8 +45,8 @@ spec:
         securityContext:
           privileged: true
         volumeMounts:
-        - name: usrbin
-          mountPath: /host/usr/bin
+        - name: localbin
+          mountPath: /host/usr/local/bin
       containers:
       - name: ovs-cni-plugin
         image: quay.io/kubevirt/ovs-cni-plugin:latest
@@ -82,9 +82,9 @@ spec:
               fieldRef:
                 fieldPath: spec.nodeName
       volumes:
-        - name: usrbin
+        - name: localbin
           hostPath:
-            path: /usr/bin
+            path: /usr/local/bin
         - name: cnibin
           hostPath:
             path: /opt/cni/bin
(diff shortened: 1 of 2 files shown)


booxter:
@phoracek please review, I believe you haven't tackled this issue yet though we had this discussion before, so I believe this is the right direction. This is needed to make ```ovs-cni``` manifests work with github.com/metalkube/dev-scripts CoreOS clusters.

booxter:
@phoracek also, while we are at it, perhaps you could clarify for me why the ```vsctl``` init container is defined for ```openshift``` manifest but not for ```kubernetes```. I thought it should be in both but perhaps I miss something.

phoracek:
@booxter thanks for the PR.

Yes, I think, we talked about this, putting the binary there sounds good, except we should update documentation too. Could you please mention in https://github.com/kubevirt/ovs-cni/blob/9115a6ca16fbda50e063429f8ceb3c8afab58856/docs/deployment-on-arbitrary-cluster.md#openshift that the `ovs-vsctl` binary is available there?

This is weird: In one of our test https://github.com/kubevirt/ovs-cni/blob/9115a6ca16fbda50e063429f8ceb3c8afab58856/tests/marker_test.go#L35 we use `ovs-vsctl` and it works even though you haven't change the `PATH`.

It is only for openshift because of we abuse the fact that OVS is already running on the node inside openshiftsdn/ovs container. We just expose access to it. There is no such image on Kubernetes. Also, there is a problem with openshift OVS container - if it is running on your host and you install and start openvswitch from RPM, it breaks openshiftsdn.

booxter:
@phoracek I am not sure which documentation change you expect. The fact that ```ovs-vsctl``` is now installed into ```/opt/cni/bin``` is an implementation detail that shouldn't bother consumers of the plugin. Right now the document says: "For OpenShift, Open vSwitch deployment is a part of OVS CNI." Considering that we assume that Open vSwitch *daemon* is provided by some other component on the host (in case of OpenShift it would be OpenShift SDN controller), I don't understand why this section is written as it is right now in the first place: The OVS CNI plugin does *not* deploy Open vSwitch *daemon*, it only deploys *ovs-vsctl* tool, and only as part of internal implementation of the plugin. With this PR merged, it becomes even more hidden from consumers. Perhaps you mean I should kill the OpenShift section from the document completely? Please clarify.

As to why tests pass calling to ```ovs-vsctl``` without absolute path or modifying ```PATH``` variable, I think it's because the provider images used for nodes install ```openvswitch``` package as part of their build process. For example, https://github.com/kubevirt/kubevirtci/blob/33b2535d25af7a68f2db007263a0b3567b363c58/k8s-multus/scripts/provision.sh#L53

phoracek:
@booxter in the documentation demo we have the whole walk-through to create a bridge on a host and connect a pod to it. With the changed bin path (why I don't expects is on PATH), user will not be able to just call `ovs-vsctl add-br brX` to the OVS database used by our CNI plugin.

You are right that "For OpenShift, Open vSwitch deployment is a part of OVS CNI." is a bit misleading, but it is about the result. If you are on OpenShift and you install OVS CNI, access to OVS database and `ovs-vsctl` is given to you. The fact that we abuse previously deployed OVS instance is just a detail (probably important when you need to debug it, but not in the default use-case).

Even if `ovs-vsctl` was installed, the OVS database socket is not on the default path - it doesn't have access to the running OVS instance.

booxter:
@phoracek Toni told me today that on CoreOS we still have ```/usr/local/bin/``` available for writing. I believe it's part of ```$PATH``` so perhaps we can just switch to using this directory instead. Thoughts?

phoracek:
@booxter Sounds good to me :) thanks

booxter:
@phoracek this should do. I've checked that it can create the files on Metalkube CoreOS cluster, and with the right contents.

booxter:
@SchSeba perhaps you can review / merge since Petr already approved.

PR #71, opened 2019-03-06
fix import path in .travis.yml script

phoracek:
In e12752a we forgot to change build script import path.


phoracek:
@SchSeba @booxter  #68 introduced an issue on travis deploy script, I believe this could fix it.

https://travis-ci.org/kubevirt/ovs-cni/builds/502455100#L647

PR #77, opened 2019-08-06
use static binding for build

phoracek:
In order to avoid issues with loading of libunbound.so.2 use
static binding.


phoracek:
@SchSeba please take a look

PR #79, opened 2019-08-28
Add mtu variable

SchSeba:
This PR add a variable to configure the mtu on the veth interface.
This feature is to make the ovs cni equal to the bridge cni plugin.

---
This change is [Reviewable](https://reviewable.io/reviews/kubevirt/ovs-cni/79)


pkg/plugin/plugin_test.go
```diff
@@ -207,7 +216,19 @@ var _ = Describe("CNI Plugin", func() {
...
+				testAddDel(conf, false, false)
+			})
+		})
+		Context("with MTU set on port", func() {
+			conf := fmt.Sprintf(`{
+				"cniVersion": "0.3.1",
+				"name": "mynet",
+				"type": "ovs",
+				"bridge": "%s",
+                "mtu": %d
```
phoracek: please fix the indentation

pkg/plugin/plugin.go
```diff
@@ -116,15 +117,15 @@ func generateRandomMac() net.HardwareAddr {
...
+func setupVeth(contNetns ns.NetNS, contIfaceName string, requestedMac string, mtu int) (*current.Interface, *current.Interface, error) {
 	hostIface := &current.Interface{}
 	contIface := &current.Interface{}
 
 	// Enter container network namespace and create veth pair inside. Doing
 	// this we will make sure that both ends of the veth pair will be removed
 	// when the container is gone.
 	err := contNetns.Do(func(hostNetns ns.NetNS) error {
-		hostVeth, containerVeth, err := ip.SetupVeth(contIfaceName, 1500, hostNetns)
+		hostVeth, containerVeth, err := ip.SetupVeth(contIfaceName, mtu, hostNetns)
```
phoracek: what if the MTU is not specified? do we set it to 0?

pkg/plugin/plugin_test.go
SchSeba: done

pkg/plugin/plugin.go
SchSeba: when it's 0 the kernel allocate a default value.

added a check for that

phoracek:
Thanks @SchSeba!

PR #85, opened 2019-11-07
bump used base images to Fedora 31

phoracek:

### DIFF
--- a/cmd/marker/Dockerfile
+++ b/cmd/marker/Dockerfile
@@ -1,4 +1,4 @@
-FROM fedora:29
+FROM fedora:31
 COPY marker /marker
 COPY .version /.version
 ENTRYPOINT [ "./marker", "-v", "3", "-logtostderr"]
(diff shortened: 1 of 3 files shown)


cmd/marker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM fedora:29
+FROM fedora:31
```
qinqon: We can even use scratch linux, for this images or even ubi8-minimal

cmd/plugin/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM fedora:29
+FROM fedora:31
```
qinqon: ditto

cmd/marker/Dockerfile
phoracek: If we do use scratch, we won't be able to use even ls to debug. We have to log to the respective node and use nsenter to do any debugging. We also won't be able to sleep. We can try to optimize it even further in another PR, but focus of this one is only to bump to a newer version.

Regarding the ubi8, it does not have openvswitch I'm any of the repositories. We need the RPM for one last call in the plugin. Once we replace the call with a socket (as Sebastian did with all the other functions), I'd be all for moving to ubi8. I opened an issue for that.

qinqon: ack

PR #89, opened 2019-11-08
use Go 1.12 on travis

phoracek:
We also need to add -mod vendor when building binaries so Go respects
the vendor/ directory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 sudo: required
 dist: trusty
-go: "1.10"
+go: "1.12"
 
 services:
 - docker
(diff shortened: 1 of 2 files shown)


fedepaol:
/lgtm

Would it make sense to try without the -mod vendor to check if the build breaks?

phoracek:
The build breaks now. CI crashes with:

```
cd cmd/marker && go fmt && go vet && GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -tags no_openssl -mod vendor
build flag -mod=vendor only valid when using modules
```

Locally it works fine.

PR #91, opened 2019-11-08
drop multus and openshift

phoracek:
There is no need to keep multus as part of this repository. It can
be easily installed via CNAO where it is properly maintained.

We should also not use OpenShift 3 which is terribly outdated. Since
our test require ovs-vsctl on host and we cannot get it to OKD hosts,
let's run tests only against Kubernetes clusters for now. Proper
integration testing happens on upper layers (CNAO).


fedepaol:
lgtm

PR #94, opened 2019-12-09
Don't crash marker on hosts without OVS

phoracek:
OVS marker should be able to live on hosts with OVS installed without
getting caught in a restart loop due to missing OVS socket.

With this change, we check if OVS is installed on the host. If it is not,
we just sleep forever. The functionality works fine, if no OVS bridge is
available, we just don't report any bridges.

Also be less graceful about failures during Update() to improve self-healing.
If our Kubernetes client is not valid anymore, it is better to restart the
marker and refresh the client.

### DIFF
--- a/cmd/marker/main.go
+++ b/cmd/marker/main.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"flag"
+	"os"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


phoracek:
@SchSeba could you please take a look?

cmd/marker/main.go
```diff
@@ -37,6 +37,11 @@ func main() {
 		glog.Fatal("ovs-socket must be set")
 	}
 
+	if _, err := os.Stat(*ovsSocket); os.IsNotExist(err) {
+		glog.Info("Given ovs-socket %q was not found. Assuming the host does not have OVS installed. Entering eternal slumber. If OVS gets installed later, please restart this pod.")
+		select {} // Wait forever
```
SchSeba: maybe create a loop with a sleep here?

phoracek: You mean to try looking for the file every X seconds? If the socket gets created on the host, will it get propagated to the mounted directory? (just if you know, otherwise, I'll try it out myself)

phoracek: Resolved, could you take a look again?

phoracek:
@qinqon would you please take a look?

phoracek:
@SchSeba ovs-socket is name of the parameter, starting with lower-case.

Thanks

PR #116, opened 2020-03-17
[WIP] Integrate QinQ functionality

rhrazdil:
Integrate OVS QinQ functionality.


docs/cni-plugin.md
```diff
@@ -35,16 +35,29 @@ Another example with a trunk port and jumbo frames:
...
 * `name` (string, required): the name of the network.
 * `type` (string, required): "ovs".
 * `bridge` (string, required): name of the bridge to use.
 * `vlan` (integer, optional): VLAN ID of attached port. Trunk port if not
-   specified.
+  specified. If CVLAN IDs are provided this VLAN ID represents SVLAN.
 * `mtu` (integer, optional): MTU.
 * `trunk` (optional): List of VLAN ID's and/or ranges of accepted VLAN
   ID's.
+* `cvlan` (optional): List of customer used VLAN ID's.
```
EdDev: You can drop the `used`.
I suggest you also explain that having this key implies that the port is using the vlan dot1q-tunnel mode.

docs/cni-plugin.md
```diff
@@ -35,16 +35,29 @@ Another example with a trunk port and jumbo frames:
...
 
+Another example with QinQ configured:
+
+```json
+{
+    "name": "myqinqnet",
+    "type": "ovs",
+    "bridge": "mynet2",
+    "vlan": 1000,
+    "cvlan": [ { "id" : 24 }, { "minID" : 100, "maxID" : 102 } ]
```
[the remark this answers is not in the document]
phoracek: I agree with the former. For the "two types" part, I'd prefer to keep it this way since it is already a format used for the trunk, let's reuse it (with all the helper functions). We can consider changes in the overall API in a follow-up.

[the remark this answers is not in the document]
rhrazdil: @EdDev @phoracek Thank you for pointing this out. Do you suggest adding additional parameter that would indicate whether to enable QinQ? Or do you think it's enough to use `cvlan` for this and allow passing empty list to include the `allow all to pass` use case?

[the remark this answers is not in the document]
phoracek: Good question.

We can introduce `vlanMode` attribute to mimic ovs-vsctl. Let's stick to the OVS database schema:

```
       VLAN Configuration:
         vlan_mode                   optional  string,  one  of  access, dot1q-tunnel, native-tagged,
                                     native-untagged, or trunk
         tag                         optional integer, in range 0 to 4,095
         trunks                      set of up to 4,096 integers, in range 0 to 4,095
         cvlans                      set of up to 4,096 integers, in range 0 to 4,095
         other_config : qinq-ethtype
                                     optional string, either 802.1ad or 802.1q
         other_config : priority-tags
                                     optional string, one of always, if-nonzero, or never
```

Details:

```
       vlan_mode:  optional  string,  one of access, dot1q-tunnel, native-tagged, native-untagged, or       trunk
              The VLAN mode of the port, as described above. When this column  is  empty,  a  default
              mode is selected as follows:

              ·      If tag contains a value, the port is an access port. The trunks column should be
                     empty.

              ·      Otherwise, the port is a trunk port. The trunks column value is honored if it is
                     present.
```

```
       cvlans: set of up to 4,096 integers, in range 0 to 4,095
              For  a dot1q-tunnel port, the customer VLANs that this port includes. If this is empty,
              the port includes all customer VLANs.

              For other kinds of ports, this setting is ignored.
```

(see man `ovs-vswitchd.conf.db`)

Our API would need some rework to fully fit this. Let's keep it simple while doing QinQ. We should require `vlanMode` to be set to `dot1q-tunnel`. If it is, we default to all cvlans or use defined range. If it is not set (or set to something else) and cvlans is set, we should return an error message.

docs/cni-plugin.md
rhrazdil: Updated, would you take another look to see if it's better?

docs/cni-plugin.md
```diff
@@ -35,16 +35,29 @@ Another example with a trunk port and jumbo frames:
...
+```
+
 ## Network Configuration Reference
 
 * `name` (string, required): the name of the network.
 * `type` (string, required): "ovs".
 * `bridge` (string, required): name of the bridge to use.
 * `vlan` (integer, optional): VLAN ID of attached port. Trunk port if not
-   specified.
+  specified. If CVLAN IDs are provided this VLAN ID represents SVLAN.
```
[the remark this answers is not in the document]
rhrazdil: Fixed

docs/cni-plugin.md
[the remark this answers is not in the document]
rhrazdil: Updated

docs/cni-plugin.md
```diff
@@ -35,16 +35,33 @@ Another example with a trunk port and jumbo frames:
...
 ```
 
+Another example with QinQ configured:
+
+```json
+{
+    "name": "myqinqnet",
+    "type": "ovs",
+    "bridge": "mynet2",
+    "vlan": 1000,
```
phoracek: Don't forget to add the vlanMode line here.

docs/cni-plugin.md
```diff
@@ -35,16 +35,33 @@ Another example with a trunk port and jumbo frames:
...
-   specified.
+  specified.
+* `vlanMode` (string, optional): Vlan mode of atached port. One of access, 
+dot1q-tunnel, native-tagged, native-untagged, or trunk. Access if not specified.
 * `mtu` (integer, optional): MTU.
 * `trunk` (optional): List of VLAN ID's and/or ranges of accepted VLAN
   ID's.
+* `cvlan` (optional): List of customer VLAN ID's for 802.1Q tunneling (QinQ).
+ If provided, VLAN ID represents service VLAN. VlanMode needs to be set to dot1q-tunnel.
+ Note that the Open vSwitch needs to have configured vlan-limit=2 in other_config.
```
phoracek: Please add this also above the QinQ config example. Include the example command line to configure that.

Similar for the vlan-limit on a bridge.

pkg/plugin/plugin_test.go
```diff
@@ -334,6 +420,47 @@ var _ = Describe("CNI Plugin", func() {
...
+				"cniVersion": "0.3.1",
+				"name": "mynet",
+				"type": "ovs",
+				"bridge": "%s",
+				"vlan": %d,
+				"vlanMode": "dot1q-tunnel"
+			}`, BRIDGE_NAME, VLAN_ID)
+			It("should successfully complete ADD and DEL commands", func() {
+				testQINQ(conf, true, "[]")
+			})
```
phoracek: Please add tests to verify error handling:

* qinq is not enabled on the ovs db
* qinq is not enabled on the bridge
* cvlan is set but the vlanMode does not match

pkg/plugin/plugin_test.go
```diff
@@ -139,6 +140,91 @@ var _ = Describe("CNI Plugin", func() {
...
+		By("Checking that port cvlan matches expected state")
+		portCvlans, err := getPortAttribute(hostIface.Name, "cvlans")
+		Expect(err).NotTo(HaveOccurred())
+		Expect(portCvlans).To(Equal(cvlans))
+
+		err = targetNs.Do(func(ns.NetNS) error {
+			defer GinkgoRecover()
+
+			for _, cvlanID := range cvlanIDs {
+				log.Printf("JSEM V LOOPE %d", cvlanID)
```
phoracek: JOOO

pkg/plugin/plugin.go
```diff
@@ -320,7 +379,16 @@ func CmdAdd(args *skel.CmdArgs) error {
 		hostIface, contIface, err = setupVeth(contNetns, args.IfName, mac, netconf.MTU)
 	}
 
-	if err = attachIfaceToBridge(ovsDriver, hostIface.Name, contIface.Name, vlanTagNum, trunks, portType, args.Netns, ovnPort); err != nil {
+	if len(cvlans) > 0 {
```
phoracek: This should not happen in usual code. Setup of vlan ifaces inside the netns is expected to be done by the user. We should whitelist those IDs for them, but they are using it. So this whole block should be dropped and used only for testing.

pkg/plugin/plugin.go
```diff
@@ -280,10 +319,30 @@ func CmdAdd(args *skel.CmdArgs) error {
...
-	portType := "access"
-	if netconf.VlanTag == nil || len(netconf.Trunk) > 0 {
-		portType = "trunk"
+	cvlans := make([]uint, 0)
+
+	if netconf.VlanMode != nil {
+		vlanMode = *netconf.VlanMode
+	}
+
+	if netconf.VlanTag != nil && netconf.Cvlan != nil {
```
phoracek: What if cvlan is not specified? It should still configure qinq, just without any limitation for cvlans (if none is set, OVS should pass all)

docs/cni-plugin.md
```diff
@@ -35,16 +35,33 @@ Another example with a trunk port and jumbo frames:
...
 ## Network Configuration Reference
 
 * `name` (string, required): the name of the network.
 * `type` (string, required): "ovs".
 * `bridge` (string, required): name of the bridge to use.
 * `vlan` (integer, optional): VLAN ID of attached port. Trunk port if not
-   specified.
+  specified.
+* `vlanMode` (string, optional): Vlan mode of atached port. One of access, 
+dot1q-tunnel, native-tagged, native-untagged, or trunk. Access if not specified.
```
EdDev: What do you mean by `Access if not specified`?
At the moment, only `trunk`, `access` and `dot1q-tunnel` are supported, right? (unsure about the native part)
Some are implicit, like `access` and `trunk`, so you can express the logic here.

I think that for this PR, you just need to support setting it to `dot1q-tunnel`, the others can be added in a separate PR.

docs/cni-plugin.md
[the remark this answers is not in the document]
EdDev: > I agree with the former. For the "two types" part, I'd prefer to keep it this way since it is already a format used for the trunk, let's reuse it (with all the helper functions). We can consider changes in the overall API in a follow-up.

I understood there is no rush with this QnQ option, therefore, I think it will be preferred to do it right from start and not change the API later on.
We can brainstorm it briefly on an API change issue. It should not take too long.

pkg/plugin/plugin.go
```diff
@@ -280,10 +319,30 @@ func CmdAdd(args *skel.CmdArgs) error {
...
+		if vlanMode != "dot1q-tunnel" {
```
pperiyasamy: what if vlanMode is not specified and cvlan is set in the NAD ? 
Let's set vlanMode to dot1q-tunnel when values are specified for cvlan.

pkg/plugin/plugin.go
```diff
@@ -280,10 +319,30 @@ func CmdAdd(args *skel.CmdArgs) error {
 	}
 
 	var vlanTagNum uint = 0
+	var vlanMode = "access"
```
EdDev: How about using constants/enums?

I also suggest you introduce `vlanMode` in a separate commit and default it there to `access` or `trunk` per the inputted keys.

docs/cni-plugin.md
rhrazdil: Right, added

docs/cni-plugin.md
rhrazdil: Added the note to the config example. Included command example.
> Similar for the vlan-limit on a bridge.

AFAIK there is no such config for bridge, is [there](http://www.openvswitch.org//ovs-vswitchd.conf.db.5.pdf)?

pkg/plugin/plugin_test.go
rhrazdil: > - qinq is not enabled on the ovs db

still TODO

> - qinq is not enabled on the bridge

As mentioned, I think there is nothing to enable on the bridge. Apart from configuring the port type, which is being tested.
> - cvlan is set but the vlanMode does not match

Added

pkg/plugin/plugin.go
rhrazdil: I see, thanks! Fixed

docs/cni-plugin.md
rhrazdil: Indeed, thanks. I rewrote the descriptions, hopefully it is more clear now.

pkg/plugin/plugin.go
```diff
@@ -173,6 +176,53 @@ func setupVeth(contNetns ns.NetNS, contIfaceName string, requestedMac string, mt
 	return hostIface, contIface, nil
 }
 
+func setupVlanLink(contNetns ns.NetNS, contIfaceName string, vlanID uint, address string) (netlink.Link, error) {
```
phoracek: This is used only by tests, hence it should be in the test module.

automation/check-patch.e2e.sh
```diff
@@ -17,6 +17,10 @@ main() {
     source automation/setup.sh
     cd ${TMP_PROJECT_PATH}
 
+    echo 'Run functional tests'
```
phoracek: Please rebase on master.

docs/cni-plugin.md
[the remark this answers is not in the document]
phoracek: Yeah, let's meet and discuss this.

PR #120, opened 2020-05-29
Bump dependencies

phoracek:
With this PR we bump dependencies to the latest release of containernetworking/ and vishvananda, Go 1.13 and Kubernetes 1.18.


go.mod
```diff
@@ -1,42 +1,50 @@
 module github.com/kubevirt/ovs-cni
 
-go 1.12
```
qinqon, line 3: go 1.13

phoracek: It is added at the bottom of the file

phoracek:
> Commit ‘ update pinned dependencies’: why did 924 file changed?

We re-fetched all libraries we depend on that are not explicitly pinned to a certain version.

> How can this be safely reviewed?

It can't. We trust tests.

PR #122, opened 2020-08-12
add test flags to make functest more readable

RamLavi:
Let's add test flags for better visability
We do that by -test.v verbose flag and E2E_TEST_ARGS flag to be able to disable ginkgo colors when running on prow.

Signed-off-by: Ram Lavi <ralavi@redhat.com>

### DIFF
--- a/hack/functests.sh
+++ b/hack/functests.sh
@@ -20,4 +20,4 @@ set -e
 
 source ./cluster/kubevirtci.sh
 
-go test ./tests --kubeconfig $(kubevirtci::kubeconfig) -ginkgo.v
+go test ./tests/... $E2E_TEST_ARGS --kubeconfig $(kubevirtci::kubeconfig)
--- a/automation/check-patch.e2e.sh
+++ b/automation/check-patch.e2e.sh
@@ -25,8 +25,7 @@ main() {
     make cluster-up
     trap teardown EXIT SIGINT SIGTERM SIGSTOP
     make cluster-sync
-    ginko_params="-ginkgo.noColor --junit-output=$ARTIFACTS_PATH/tests.junit.xml"
-    FUNC_TEST_ARGS=$ginko_params make functest
+    make E2E_TEST_ARGS="-ginkgo.v -test.v -ginkgo.noColor --junit-output=$ARTIFACTS/junit.functest.xml" functest
 }
 
 [[ "${BASH_SOURCE[0]}" == "$0" ]] && main "$@"


RamLavi:
@phoracek do you think I should also add polarion flags `--test-suite-params="$POLARION_TEST_SUITE_PARAMS"` and ginkgo e2e test args `E2E_TEST_ARGS`?

hack/functests.sh
```diff
@@ -20,4 +20,4 @@ set -e
 
 source ./cluster/kubevirtci.sh
 
-go test ./tests --kubeconfig $(kubevirtci::kubeconfig) -ginkgo.v
+go test ./tests/... --kubeconfig $(kubevirtci::kubeconfig) -ginkgo.v -test.v -race
```
phoracek, line 23: Do we want to run our e2e tests with -race? I would expect it for unit tests, but anything above it sounds like a potential source of flakes. Please keep this PR to the point of "more readable". We can decide to make it -race later.

phoracek:
OVS CNI is not linked in polarion, so we don't need that flag.

E2E args would be useful to play with ginkgo arguments. e.g. we may want to disable colors on Prow.

hack/functests.sh
RamLavi: removed `-race`.
Done

hack/functests.sh
```diff
@@ -20,4 +20,4 @@ set -e
 
 source ./cluster/kubevirtci.sh
 
-go test ./tests --kubeconfig $(kubevirtci::kubeconfig) -ginkgo.v
+go test ./tests/... $E2E_TEST_ARGS --kubeconfig $(kubevirtci::kubeconfig) -ginkgo.v -test.v
```
phoracek, line 23: Why don't wee pass -v through E2E_TEST_ARGS? It should be up to the user to enable verbosity

RamLavi: I think the verbose is important enough to be in the default. 
Would you prefer I removed it from here and added it as `E2E_TEST_ARGS="-ginkgo.v -test.v"` in the `functest` make recepie?

phoracek: The issue with verbose is that it spams the output with passing tests, but we are mostly interested in those that fails. So for CI run, dumping all is informative about progress at least. When running locally, I don't care about old passing tests.

go.mod
```diff
@@ -19,7 +19,7 @@ require (
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0 // indirect
 	k8s.io/apimachinery v0.18.3
 	k8s.io/client-go v0.0.0-00010101000000-000000000000
-	kubevirt.io/qe-tools v0.1.3
+	kubevirt.io/qe-tools v0.1.6
```
phoracek, line 22: Wrong rebase?

RamLavi: yep, made a mess :D, fix shortly

PR #123, opened 2020-08-12
add multus to cluster-up deployment

RamLavi:
In order to use multus-cni in tests, multus must be deployed in the cluster.
Let's add this as part of the cluster-up sequence.

Signed-off-by: Ram Lavi <ralavi@redhat.com>


RamLavi:
@phoracek another option is to deploy cnao with multus. This has the advantage of keeping future tests more stable.
On the other hand, using latest multus yaml from master may find broken issues sooner. 
what do you think?

phoracek:
I would start with what you presented here, just for the sake of simplicity. If we figure out it is problematic, we can upgrade to CNAO

PR #124, opened 2020-08-13
expand ovs tests

RamLavi:
We want to expand ovs-cni coverage to also do tests on k8s cluster.

- **ovs-test, Fix pods exists test**
Currently It(assert pods exists) test checks if there is any pod,
to check if ovs-cni pod is up and running.
Since there is already a default pod named local-volume-provisioner in the cluster,
we need to make this test more specific to catch ovs-cni pod.
we also change from It() to Specify for better readability

- **ovs-test, Add utils function helpers**
We need to add cluster, node and pod test packages.
These function helpers will be used in ovs cluster tests.

- **ovs-test, Add ovs-cni l2 connectivity tests between 2 pods**
Currently there is only unit test checking ovs-cni is properly working.
added a new test that will check connectivity between 2 pods connected by the host's ovs-bridge.

- **ovs-test, Add go mod and auto-generated vendor folders**
bunched all go mod generated files for easier reviewing


RamLavi:
after rebasing merged PR #123 e2e test should pass

tests/utils.go
```diff
@@ -20,7 +20,13 @@
 package tests
```
ormergi, line 21: Please change this file name to something like :pod_utils so it will be utils for pods only and not helpers junkyard :D

tests/tests.go
```diff
@@ -18,7 +18,128 @@
...
+			Namespace:   testNamespace,
+			Annotations: map[string]string{"k8s.v1.cni.cncf.io/resourceName": "ovs-cni.network.kubevirt.io/" + bridgeName },
+		},
+		Spec: netv1.NetworkAttachmentDefinitionSpec{
+			Config: config,
+		},
+	}
+
+	_, err := netClient.NetworkAttachmentDefinitions(nad.Namespace).Create(context.TODO(), nad, metav1.CreateOptions{})
+	Expect(err).ToNot(HaveOccurred(), "should succeed creating nad NetworkAttachmentDefinition")
```
ormergi, line 122: I am not sure the error message will be informative, will it show there error message from the `Create` command?

tests/tests.go
```diff
@@ -18,7 +18,128 @@
...
+	}
+}
+
+func createNetworkAttachmentDefinition(nadName, bridgeName, config string) {
+	By(fmt.Sprintf("Adding NetworkAttachmentDefinition %s of ovs bridge", nadName))
+	nad := &netv1.NetworkAttachmentDefinition{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        nadName,
+			Namespace:   testNamespace,
+			Annotations: map[string]string{"k8s.v1.cni.cncf.io/resourceName": "ovs-cni.network.kubevirt.io/" + bridgeName },
```
ormergi, line 114: Is it mandatory to add this annotation?

tests/ovs_test.go
```diff
@@ -32,4 +35,42 @@ var _ = Describe("ovs-cni", func() {
...
+				})
+				AfterEach( func() {
+					removeNetworkAttachmentDefinition(nadName)
+					deletePodsInTestNamespace()
+				})
+
+				It("Should enable connectivity between two pods connected to the same bridge", func() {
+					cidrPod1 := "11.0.0.1/24"
+					cidrPod2 := "11.0.0.2/24"
+					createPrivilegedPodWithIp("pod-test-1", nadName, cidrPod1)
```
ormergi, line 63: Why do we need to create privileged pod for pinging, non-privileged one wont be enough?

tests/ovs_test.go
```diff
@@ -32,4 +35,42 @@ var _ = Describe("ovs-cni", func() {
 			})
 		})
 	})
+
+	Describe("pod ovs-bridge connectivity tests", func() {
+		bridgeName := "br-test"
+		Context("When creating ovs-bridge on the host node", func() {
+			BeforeEach( func() {
```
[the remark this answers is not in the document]
RamLavi: If I understood correctly @phoracek's former reviews on the matter, what is said in the `Context` text needs to be set in the immediate `BeforeEach` that follows it.

tests/utils.go
RamLavi: This file has 2 functions: 1 performs ssh to node, and the other exec to pod.
Is that alright if I split them to `pod_utils.go` and `node_utils.go`?

tests/tests.go
RamLavi: yes it will present it if err is not nil

tests/tests.go
RamLavi: the purpose of the this annotation is to make sure the pod is assigned to a node that has this ovs bridge. 
Since we have only 1 node, it is not really necessary. do you think I should remove it?

tests/ovs_test.go
RamLavi: unfortunately not, since we need to manually change the pod's ip (ovs-cni configuration does not currently handle ipam configuration).

tests/ovs_test.go
```diff
@@ -32,4 +35,42 @@ var _ = Describe("ovs-cni", func() {
...
+				BeforeEach( func() {
+					createNetworkAttachmentDefinition(nadName, bridgeName, `{ "cniVersion": "0.3.1", "type": "ovs", "bridge": "` + bridgeName + `", "vlan": 100 }`)
+				})
+				AfterEach( func() {
+					removeNetworkAttachmentDefinition(nadName)
+					deletePodsInTestNamespace()
+				})
+
+				It("Should enable connectivity between two pods connected to the same bridge", func() {
+					cidrPod1 := "11.0.0.1/24"
```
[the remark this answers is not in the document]
RamLavi: Done

tests/ovs_test.go
```diff
@@ -32,4 +35,42 @@ var _ = Describe("ovs-cni", func() {
...
+					cidrPod2 := "11.0.0.2/24"
+					createPrivilegedPodWithIp("pod-test-1", nadName, cidrPod1)
+					createPrivilegedPodWithIp("pod-test-2", nadName, cidrPod2)
+
+					By("Checking pods connectivity by pinging from one to the other")
+					ipPod2, _, err := net.ParseCIDR(cidrPod2)
+					Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("should succeed parsing pod's cidr: %s", cidrPod2))
+					out, _, err := ExecToPod("pod-test-1", "test", "ping -c 1 " + ipPod2.String())
```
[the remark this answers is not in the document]
RamLavi: Done

tests/ovs_test.go
```diff
@@ -32,4 +35,42 @@ var _ = Describe("ovs-cni", func() {
...
+					Expect(err).ToNot(HaveOccurred(), "should succeed performing ping command with no error")
```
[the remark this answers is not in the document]
RamLavi: Done

RamLavi:
> About commit "add ovs-cni l2 connectivity tests between 2 pods.":
> You added few functions that related to creating and deleting cluster objects, could you please split it to another commit?
> also I am not sure `test/tests.go` is the right place, maybe `tests/cluster_utils.go` will be better fit?
I will add it in another commit, with new name `tests/cluster_utils.go`
Done
> 
> Please consider extracting all `Expect` usage in those functions out, so usage of `Expect` will be only from the tests itself and not inside a helper function.
Done
> 
> About commit "adding go mod and vendor folders":
> Do we need vendor directory if we use go modules?
vendor is the product of `go mod tidy vendor`, so we do need it.

tests/utils.go
RamLavi: Done

tests/tests.go
ormergi: Ok I understand now, it is good to have it maybe it worth adding comment about it

tests/ovs_test.go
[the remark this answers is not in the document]
ormergi: Sorry for the nose, I though I saw something else

tests/ovs_test.go
ormergi: Why is it necessary to set specific IP address for the pod?

tests/ovs_test.go
```diff
@@ -27,8 +27,8 @@ var _ = Describe("ovs-cni", func() {
 	Describe("pod availability tests", func() {
 		Context("pod availability tests", func() {
 			It("assert pods exists", func() {
-				pods, _ := clientset.CoreV1().Pods("").List(context.TODO(), v1.ListOptions{})
-				Expect(len(pods.Items)).Should(BeNumerically(">", 0))
+				pods, _ := clientset.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{ LabelSelector: "app=ovs-cni"})
```
maiqueb, line 33: does this pass the syntax check ? (extra space in `ListOptions{-> <-LabelSelector: "app=ovs-cni"})`

ormergi:
> > About commit "add ovs-cni l2 connectivity tests between 2 pods.":
> > You added few functions that related to creating and deleting cluster objects, could you please split it to another commit?
> > also I am not sure `test/tests.go` is the right place, maybe `tests/cluster_utils.go` will be better fit?
> > I will add it in another commit, with new name `tests/cluster_utils.go`
> 
> Done
> 
> > Please consider extracting all `Expect` usage in those functions out, so usage of `Expect` will be only from the tests itself and not inside a helper function.
> 
> Done
> 
> > About commit "adding go mod and vendor folders":
> > Do we need vendor directory if we use go modules?
> 
> vendor is the product of `go mod tidy vendor`, so we do need it.

Thanks

tests/cluster_utils.go
```diff
@@ -0,0 +1,167 @@
...
+	*privileged = true
+	podObject := &corev1.Pod{ObjectMeta: metav1.ObjectMeta{
+		Name: podName,
+		Namespace: testNamespace,
+		Annotations: map[string]string{"k8s.v1.cni.cncf.io/networks": nadName},
+	},
+	Spec: corev1.PodSpec{ Containers: []corev1.Container{{Name: "test",
+		Image:   "alpine",
+		Command: []string{"sh", "-c", fmt.Sprintf("ip address add %s dev net1; sleep 99999", cidr)},
+		SecurityContext: &corev1.SecurityContext{ Privileged: privileged}}}}}
```
maiqueb, line 77: Shouldn't this be formatted ? I find it misleading to see it as is.

tests/cluster_utils.go
```diff
@@ -0,0 +1,167 @@
...
+	privileged := new(bool)
+	*privileged = true
+	podObject := &corev1.Pod{ObjectMeta: metav1.ObjectMeta{
+		Name: podName,
+		Namespace: testNamespace,
+		Annotations: map[string]string{"k8s.v1.cni.cncf.io/networks": nadName},
+	},
+	Spec: corev1.PodSpec{ Containers: []corev1.Container{{Name: "test",
+		Image:   "alpine",
+		Command: []string{"sh", "-c", fmt.Sprintf("ip address add %s dev net1; sleep 99999", cidr)},
```
maiqueb, line 76: Isn't there a 'smarter' way to assign an IP address to the pod  on alpine ? We've stumbled on this particular rock plenty of times already.

tests/pod_utils.go
```diff
@@ -0,0 +1,71 @@
...
+	"bytes"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/remotecommand"
+	"github.com/pkg/errors"
+)
+
+func execToPod(podName, containerName, command string) (string, string, error) {
```
maiqueb, line 32: shouldn't this be named `execOn...` ?

tests/cluster_utils.go
```diff
@@ -0,0 +1,167 @@
...
+func removeTestNamespaces() {
+	By(fmt.Sprintf("Waiting for namespace %s to be removed, this can take a while ...\n", testNamespace))
+	EventuallyWithOffset(1, func() bool { return apierrors.IsNotFound(deleteNamespace(testNamespace)) }, 120*time.Second, 5*time.Second).
+		Should(BeTrue(), "Namespace %s haven't been deleted within the given timeout", testNamespace)
+}
+
+func createPrivilegedPodWithIp(podName, nadName, cidr string) error {
+	By(fmt.Sprintf("Creating pod %s with priviliged premission and ip %s", podName, cidr))
+	privileged := new(bool)
+	*privileged = true
```
maiqueb, line 68: you could just do `privileged := true` then use `&privileged`; that would save you the constructor line.

tests/ovs_test.go
```diff
@@ -32,4 +35,54 @@ var _ = Describe("ovs-cni", func() {
 			})
 		})
 	})
+
+	Describe("pod ovs-bridge connectivity tests", func() {
+		bridgeName := "br-test"
+		Context("When creating ovs-bridge on the host node", func() {
+			BeforeEach( func() {
+				err := removeOvsBridgeOnNode(bridgeName)
```
maiqueb, line 43: I don't quite remember where we are with this, but I **do** remember a strong push from @EdDev to stop removing stuff in beforeEach's - each test should make sure whatever it created is removed. 

Since you're pretty much starting this module from scratch, I think that would be a good principle to use (if you have leaked stuff, let's not place a rug on top of it).

RamLavi:
> I saw some syntax related stuff, which leads me to think we don't enforce syntax check on ovs-cni.
> 
> While I think that's far from good, it's out of scope of this PR; you can disregard my comments about it.

Do you mean something like [macvtap's](https://github.com/kubevirt/macvtap-cni/blob/master/Makefile#L41) `make fmt`? 
If indeed, I will add it in a different PR

tests/ovs_test.go
RamLavi: Done

tests/cluster_utils.go
RamLavi: Done

tests/pod_utils.go
RamLavi: Done

tests/cluster_utils.go
RamLavi: true. Done

maiqueb:
> > I saw some syntax related stuff, which leads me to think we don't enforce syntax check on ovs-cni.
> > While I think that's far from good, it's out of scope of this PR; you can disregard my comments about it.
> 
> Do you mean something like [macvtap's](https://github.com/kubevirt/macvtap-cni/blob/master/Makefile#L41) `make fmt`?
> If indeed, I will add it in a different PR

Yes, something like that.

> 
> All the things you noticed - I will manually fix

Thanks for listening.

tests/ovs_test.go
RamLavi: I get what you mean. The purpose of adding this for the scenario where I stop the test manually so that the `AfterEach` is skipped, and so we can be left the ovs-bridge for the next test.
But I agree it's not enough of a reason, and it could put things under the rug. will remove.
Done

ormergi:
What you think about following the principal of creating module for each utilities module, by creating dedicated directory and file, like @EdDev  presented in kubevirt'/kubevirt:
https://github.com/kubevirt/kubevirt/tree/master/tests/containerdisk
then the usage will look like:
https://github.com/kubevirt/kubevirt/blob/master/tests/stability_test.go#L9
https://github.com/kubevirt/kubevirt/blob/master/tests/stability_test.go#L26

tests/cluster_utils.go
RamLavi: I'm afraid I don't know of such a way. can you direct me to somewhere it is done?

RamLavi:
@maiqueb [fixed](https://github.com/kubevirt/ovs-cni/compare/c40e59d8fc1bca7f1bbeb655dfe47abaaca2983a..c69c2267835adef138105977c770af3181b9ad1a) your comments.
Regarding configuring the Alpine test-pod, I didn't find a way to do it otherwise. can you point me to an example?

tests/cluster_utils.go
ormergi: Since we need to configure a secondary NIC and OVS does not support IPAM configuration, you have to do it "form inside the pod"
If it was we could configure it with calico https://docs.projectcalico.org/networking/use-specific-ip#how-to

tests/tests.go
RamLavi: Done

tests/ovs_test.go
RamLavi: simply easier to ping if you know the ips..

tests/ovs_test.go
```diff
@@ -27,8 +27,8 @@ var _ = Describe("ovs-cni", func() {
 	Describe("pod availability tests", func() {
 		Context("pod availability tests", func() {
 			It("assert pods exists", func() {
-				pods, _ := clientset.CoreV1().Pods("").List(context.TODO(), v1.ListOptions{})
-				Expect(len(pods.Items)).Should(BeNumerically(">", 0))
+				pods, _ := clientset.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{ LabelSelector: "app=ovs-cni"})
+				Expect(len(pods.Items)).Should(BeNumerically(">", 0), "should have at least 1 ovs-cni pod deployed")
```
alonSadan, line 34: nit: maybe s/`Should`/`To` ? 
`Expect(len(pods.Items)).To(BeNumerically(">", 0), "should have at least 1 ovs-cni pod deployed")`

tests/cluster_utils.go
```diff
@@ -0,0 +1,166 @@
...
+	By(fmt.Sprintf("creating %s namespace", testNamespace))
+	_, err := clientset.CoreV1().Namespaces().Create(context.TODO(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: testNamespace}}, metav1.CreateOptions{})
+	if err != nil {
+		return errors.Wrap(err, "failed to create test namespace")
+	}
+	return nil
+}
+
+func deleteNamespace(namespace string) error {
+	return clientset.CoreV1().Namespaces().Delete(context.TODO(), namespace, metav1.DeleteOptions{})
```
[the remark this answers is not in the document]
RamLavi: it is actually. `deleteNamespace` is a private function that is consumed by `RemoveTestNamespace`..
`RemoveTestNamespace` does an `Eventually` in order to make sure the namespace is terminated appropriately.

this is why in the private function we don't need the Wrap.

In general, I add Wrap to all helper functions as it increases readability in case it fails there.

tests/cluster_utils.go
```diff
@@ -0,0 +1,166 @@
...
+	netv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+)
+
+var kubeconfig *string
+var restConfig *restclient.Config
+var clientset *kubernetes.Clientset
+var netClient *netclient.K8sCniCncfIoV1Client
+const testNamespace = "test-namespace"
+
+func createTestNamespaces() error {
```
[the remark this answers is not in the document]
RamLavi: Done

tests/cluster_utils.go
```diff
@@ -0,0 +1,166 @@
...
+		return errors.Wrap(err, "failed to create test namespace")
+	}
+	return nil
+}
+
+func deleteNamespace(namespace string) error {
+	return clientset.CoreV1().Namespaces().Delete(context.TODO(), namespace, metav1.DeleteOptions{})
+}
+
+func removeTestNamespaces() {
```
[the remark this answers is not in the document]
RamLavi: Done

tests/cluster_utils.go
```diff
@@ -0,0 +1,166 @@
...
+	By(fmt.Sprintf("Waiting for namespace %s to be removed, this can take a while ...\n", testNamespace))
+	EventuallyWithOffset(1, func() bool { return apierrors.IsNotFound(deleteNamespace(testNamespace)) }, 120*time.Second, 5*time.Second).
```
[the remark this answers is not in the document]
RamLavi: cool! Done

tests/cluster_utils.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to get pod list in namespace %s", testNamespace)
+	}
+
+	for _, pod := range podList.Items {
+		err = clientset.CoreV1().Pods(testNamespace).Delete(context.TODO(), pod.Name, metav1.DeleteOptions{})
+		if err != nil {
+			return errors.Wrapf(err, "failed to delete pod %s", pod.Name)
+		}
+	}
```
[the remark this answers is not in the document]
RamLavi: good catch! 
Done

tests/cluster_utils.go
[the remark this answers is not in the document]
alonSadan: ah I see.
Thanks for explaining.

tests/ovs_test.go
RamLavi: Thank you. I'm not sure switching Should-->To will make it more readable..
If for example I induce the Expect to fail, it will look like this:
```
• Failure [0.007 seconds]
ovs-cni
/home/ralavi/go/src/github.com/kubevirt/ovs-cni/tests/ovs_test.go:32
  pod availability tests
  /home/ralavi/go/src/github.com/kubevirt/ovs-cni/tests/ovs_test.go:33
    pod availability tests
    /home/ralavi/go/src/github.com/kubevirt/ovs-cni/tests/ovs_test.go:34
      assert pods exists [It]
      /home/ralavi/go/src/github.com/kubevirt/ovs-cni/tests/ovs_test.go:35

      should have at least 1 ovs-cni pod deployed
      Expected
          <int>: 1
      to be <
          <int>: 0

      /home/ralavi/go/src/github.com/kubevirt/ovs-cni/tests/ovs_test.go:37
```

if it would read as `To have at least 1 ovs-cni pod deployed` - I think it'll be less understood. 
But maybe I can try to make the Context more readable.

RamLavi: Done

RamLavi:
> Thanks Ram. reviewed the first three commits. looks good:).
> I added some suggestions and nits.

you can review the fixes: [here](https://github.com/kubevirt/ovs-cni/compare/c69c2267835adef138105977c770af3181b9ad1a..70a86a1e4126016e0cc0d2c2ba864ecf84b3278f) and also [here](https://github.com/kubevirt/ovs-cni/compare/70a86a1e4126016e0cc0d2c2ba864ecf84b3278f..0b1b5fcc56dc59986b303825a654375a29cf1c25)

RamLavi:
> What you think about following the principal of creating module for each utilities module, by creating dedicated directory and file, like @EdDev presented in kubevirt'/kubevirt:
> https://github.com/kubevirt/kubevirt/tree/master/tests/containerdisk
> then the usage will look like:
> https://github.com/kubevirt/kubevirt/blob/master/tests/stability_test.go#L9
> https://github.com/kubevirt/kubevirt/blob/master/tests/stability_test.go#L26
> 
> So in our case, for example:
> There will be a file `tests/cluster/cluster.go` and we could use it in tests like this
> 
> ```
> import github.com/kubevirt/ovs-cni/tests/cluster
> ...
> cluster.CreateTestsNamespace...
> ```

fix introduced in this [push](https://github.com/kubevirt/ovs-cni/compare/c69c2267835adef138105977c770af3181b9ad1a..70a86a1e4126016e0cc0d2c2ba864ecf84b3278f)

tests/ovs_test.go
```diff
@@ -35,4 +38,52 @@ var _ = Describe("ovs-cni", func() {
 			})
 		})
 	})
+
+	Describe("pod ovs-bridge connectivity tests", func() {
+		bridgeName := "br-test"
+		Context("When creating ovs-bridge on the host node", func() {
+			BeforeEach( func() {
+				err := cluster.AddOvsBridgeOnNode(bridgeName)
+				ExpectWithOffset(1, err).ToNot(HaveOccurred(), "should succeed adding ovs bridge on node")
```
alonSadan, line 47: can you explain why to use `ExpectWithOffset` in the outer context as well (or at all)?

tests/ovs_test.go
```diff
@@ -35,4 +38,52 @@ var _ = Describe("ovs-cni", func() {
...
+				AfterEach( func() {
+					err := cluster.RemoveNetworkAttachmentDefinition(nadName)
+					ExpectWithOffset(1, err).ToNot(HaveOccurred(), "should succeed deleting NetworkAttachmentDefinition")
+
+					err = cluster.DeletePodsInTestNamespace()
+					ExpectWithOffset(1, err).ToNot(HaveOccurred(), "should succeed deleting pods under test namespace")
+				})
+
+				It("Should enable connectivity between two pods connected to the same bridge", func() {
+					pod1Name := "pod-test-1"
```
alonSadan, line 69: consider using const 
```
const (pod1Name = "pod-test-1"
           pod2Name = "pod-test-2"
           cidrPod1 = "10.0.0.1/24"
           cidrPod2 = "10.0.0.2/24"
)
```

tests/cluster_utils.go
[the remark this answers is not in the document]
alonSadan: sorry, I know this is  something I suggested, but maybe 6 min is too much?
what do you think?

tests/ovs_test.go
RamLavi: in this test suite I have 3 modules (cluster, node, pod) with helper functions.
Following the talk with @EdDev about how make tests better and more readable, we said that the helper should not use Expect, and instead return an error (preferably with Wrap). 
`ExpectWithOffset` is like regular `Expect` but it refers to the last function (where 1 the stack offset, since I want the Expect to refer  to `AddOvsBridgeOnNode` and not a more internal function in it)

Using `ExpectWithOffset`, I can avoid using `Expect` on the helper functions, and still make sure the the Expect relates to them.

alonSadan: Got it.  Thanks

tests/cluster_utils.go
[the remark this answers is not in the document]
RamLavi: I'm not sure. I noticed pods do take relatively long time to be deleted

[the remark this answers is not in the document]
alonSadan: I see that in kubevirt/kubevirt , at least  [here](https://github.com/kubevirt/kubevirt/blob/master/tests/storage_test.go#L105) and [here](https://github.com/kubevirt/kubevirt/blob/master/tests/migration_test.go#L1011) they are waiting for 3 minutes for one pod to be deleted. so maybe 6 min is OK

tests/ovs_test.go
RamLavi: Done

tests/cluster/cluster.go
```diff
@@ -0,0 +1,181 @@
...
+		Command: []string{"sh", "-c", fmt.Sprintf("ip address add %s dev net1; sleep 99999", cidr)},
+		SecurityContext: &corev1.SecurityContext{Privileged: &privileged}}}}}
+
+	_, err := Clientset.CoreV1().Pods(TestNamespace).Create(context.TODO(), podObject, metav1.CreateOptions{})
+	if err != nil {
+		return errors.Wrap(err, "failed to wait for pod container to be in Running state")
+	}
+
+	By("waiting for pod container to be in Running state")
+	Eventually(func() bool {
```
phoracek, line 92: If this is used only in BeforeEach, we don't need to return issues and stick to `Eventually` and `Expect`. I believe that would be valid.

If there is an util that is used within `It`, it should not use `Gomega` asserters and return an error instead.

tests/cluster/cluster.go
```diff
@@ -0,0 +1,181 @@
...
+		if err != nil {
+			return errors.Wrapf(err, "failed to delete pod %s", pod.Name)
+		}
+	}
+
+	Eventually(func() int {
+		podsList, err := Clientset.CoreV1().Pods(TestNamespace).List(context.TODO(), metav1.ListOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		return len(podsList.Items)
+	}, 6*time.Minute, 1*time.Second).Should(BeZero())
```
phoracek, line 131: No need for `1*` if it is just a second.

tests/cluster/cluster.go
```diff
@@ -0,0 +1,181 @@
...
+	By("Cleaning NetworkAttachmentDefinition")
+	err := NetClient.NetworkAttachmentDefinitions(TestNamespace).Delete(context.TODO(), nadName, metav1.DeleteOptions{})
+	if err != nil {
+		return errors.Wrap(err, "failed to delete nad NetworkAttachmentDefinition")
+	}
+	return nil
+}
+
+func RemoveOvsBridgeOnNode(bridgeName string) error {
+	By("removing ovs-bridge on the node")
```
phoracek, line 166: Capital

tests/ovs_test.go
```diff
@@ -20,15 +20,15 @@ import (
...
 )
 
 var _ = Describe("ovs-cni", func() {
 	Describe("pod availability tests", func() {
-		Context("pod availability tests", func() {
-			It("assert pods exists", func() {
-				pods, _ := clientset.CoreV1().Pods("").List(context.TODO(), v1.ListOptions{})
-				Expect(len(pods.Items)).Should(BeNumerically(">", 0))
+		Context("When ovs-cni is deployed on the cluster", func() {
+			It("should assert if ovs-cni pod doesn't exist", func() {
```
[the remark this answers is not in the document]
RamLavi: Done

tests/cluster/cluster.go
```diff
@@ -0,0 +1,181 @@
...
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	netclient "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/clientset/versioned/typed/k8s.cni.cncf.io/v1"
+	netv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+
+	"github.com/kubevirt/ovs-cni/tests/node"
+)
+
+var Kubeconfig *string
```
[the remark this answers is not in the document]
RamLavi: I was sure we need it here but in fact we don't. moved to private inside test_suite.go.
Done

tests/cluster/cluster.go
```diff
@@ -0,0 +1,181 @@
...
+var RestConfig *restclient.Config
+var Clientset *kubernetes.Clientset
+var NetClient *netclient.K8sCniCncfIoV1Client
+var TestNamespace = "test-namespace"
+
+func CreateTestNamespace() error {
+	By(fmt.Sprintf("creating %s namespace", TestNamespace))
```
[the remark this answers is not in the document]
RamLavi: Done

tests/cluster/cluster.go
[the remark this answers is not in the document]
phoracek: In fact, we should try to avoid any globals, especially in a utils lib. Would you create a helper function initializing a needed client for each user?

tests/cluster/cluster.go
```diff
@@ -0,0 +1,181 @@
...
+	netv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+
+	"github.com/kubevirt/ovs-cni/tests/node"
+)
+
+var Kubeconfig *string
+var RestConfig *restclient.Config
+var Clientset *kubernetes.Clientset
+var NetClient *netclient.K8sCniCncfIoV1Client
+var TestNamespace = "test-namespace"
```
phoracek, line 46: Could be a const?

tests/cluster/cluster.go
```diff
@@ -0,0 +1,181 @@
...
+
+func CreateTestNamespace() error {
+	By(fmt.Sprintf("creating %s namespace", TestNamespace))
+	_, err := Clientset.CoreV1().Namespaces().Create(context.TODO(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: TestNamespace}}, metav1.CreateOptions{})
+	if err != nil {
```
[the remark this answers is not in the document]
RamLavi: Done

tests/cluster/cluster.go
```diff
@@ -0,0 +1,181 @@
...
+		return errors.Wrap(err, "failed to create test namespace")
+	}
+	return nil
+}
+
+func RemoveTestNamespace() error {
+	By(fmt.Sprintf("Waiting for namespace %s to be removed, this can take a while ...\n", TestNamespace))
```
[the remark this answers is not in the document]
RamLavi: Done

tests/cluster/cluster.go
RamLavi: ack.
fixed throughout the test.
Done

tests/cluster/cluster.go
RamLavi: Done

tests/cluster/cluster.go
RamLavi: Done

tests/marker_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package tests_test
```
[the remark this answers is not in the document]
RamLavi: Done

tests/cluster/cluster.go
RamLavi: const is  local, and I need TestNamespace in the `pod.PingFromPod`

moving `PingFromPod` to cluster module (where TestNamespace lives) wouldn't work also since it'll create import loop, so I left it as that.

phoracek: Const should be public too as long as they start with a capital.

https://github.com/kubevirt/cluster-network-addons-operator/blob/master/pkg/names/names.go

tests/cluster/cluster.go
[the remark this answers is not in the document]
RamLavi: isn't keeping it private global in `tests_suite_test.go` safer?

RamLavi:
> Review of the first couple of commits

[fixed](https://github.com/kubevirt/ovs-cni/compare/e042a2cc0652a741207502452d1aa177c071ab9d..811d4d0403d74c0d9a7e9ba9def1458b69390467)

tests/cluster/cluster.go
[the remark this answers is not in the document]
phoracek: No, utils are imported by  the test suite, they must not depend on a global kept in it.

For utils, clients should be initialized inside helper functions OR passed down as an attribute, but that makes it unnecessarily verbose.

For tests that need a client, you can use a global initialized in the `BeforeSuite`

tests/cluster/cluster.go
```diff
@@ -0,0 +1,152 @@
...
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	netclient "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/clientset/versioned/typed/k8s.cni.cncf.io/v1"
+	netv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+
+	"github.com/kubevirt/ovs-cni/tests/node"
+)
+
+var RestConfig *restclient.Config
```
phoracek, line 42: Sorry I wasn't clear. My comment to avoid globals was for all that are listed here.

tests/ovs_test.go
```diff
@@ -35,4 +38,52 @@ var _ = Describe("ovs-cni", func() {
 			})
 		})
 	})
+
+	Describe("pod ovs-bridge connectivity tests", func() {
+		const bridgeName = "br-test"
```
phoracek, line 43: Please keep the definition in its relevant context. So inside the following Context

tests/ovs_test.go
```diff
@@ -35,4 +38,52 @@ var _ = Describe("ovs-cni", func() {
...
+		Context("When creating ovs-bridge on the host node", func() {
```
phoracek, line 44: We should use past tense. The test inside is not running while we are creating the bridge, it is running "when an OVS bridge is configured on a node"

tests/ovs_test.go
```diff
@@ -35,4 +38,52 @@ var _ = Describe("ovs-cni", func() {
...
+			BeforeEach( func() {
+				cluster.AddOvsBridgeOnNode(bridgeName)
+			})
+			AfterEach( func() {
+				cluster.RemoveOvsBridgeOnNode(bridgeName)
+			})
+
+			Context("and adding ovs network with vlan", func() {
```
phoracek, line 52: "and a network attachment definition is defined"

tests/ovs_test.go
```diff
@@ -35,4 +38,52 @@ var _ = Describe("ovs-cni", func() {
...
+				const nadName = "ovs-net"
+				BeforeEach( func() {
+					cluster.CreateNetworkAttachmentDefinition(nadName, bridgeName, `{ "cniVersion": "0.3.1", "type": "ovs", "bridge": "` + bridgeName + `", "vlan": 100 }`)
+				})
+				AfterEach( func() {
+					cluster.RemoveNetworkAttachmentDefinition(nadName)
+				})
+
+				Context("and deploying two pods in the node holding the ovs-bridge", func() {
```
phoracek, line 61: "and two pods are connected through it"

tests/ovs_test.go
```diff
@@ -35,4 +38,52 @@ var _ = Describe("ovs-cni", func() {
...
+					)
+					BeforeEach( func() {
+						cluster.CreatePrivilegedPodWithIp(pod1Name, nadName, cidrPod1)
+						cluster.CreatePrivilegedPodWithIp(pod2Name, nadName, cidrPod2)
+					})
+					AfterEach( func() {
+						cluster.DeletePodsInTestNamespace()
+					})
+
+					It("Should enable connectivity between two pods connected to the same bridge", func() {
```
phoracek, line 76: Specify "they should be able to communicate over the network"

tests/ovs_test.go
```diff
@@ -35,4 +38,52 @@ var _ = Describe("ovs-cni", func() {
...
+						By("Checking pods connectivity by pinging from one to the other")
+						ipPod2, _, err := net.ParseCIDR(cidrPod2)
+						Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("should succeed parsing pod's cidr: %s", cidrPod2))
+
+						out, err := pod.PingFromPod(pod1Name, "test", ipPod2.String())
+						ExpectWithOffset(1, err).ToNot(HaveOccurred(), fmt.Sprintf("should be able to ping from pod %s to IP %s which is on pod %s", pod1Name, ipPod2, pod2Name))
```
phoracek, line 82: We don't need WithOffset when we are in a test

tests/ovs_test.go
```diff
@@ -35,4 +38,52 @@ var _ = Describe("ovs-cni", func() {
...
+						Expect(out).To(ContainSubstring("0% packet loss"), "Should not have any packet loss")
```
phoracek, line 83: If you have the ping helper running with -w 3 -c 1, it should return fail when unable to ping. In case of success we don't need to record the output and simply check for `ToNot(HaveOccurred)`

tests/tests_suite_test.go
```diff
@@ -45,11 +46,20 @@ var _ = BeforeSuite(func() {
 
 	var err error
 	cluster.RestConfig, err = clientcmd.BuildConfigFromFlags("", *kubeconfig)
```
phoracek, line 48: Define these as globals of this suite, so we are not changing our dependencies.

phoracek (changes requested):

tests/tests_suite_test.go
RamLavi: I actually need `RestConfig` in `node.execOnPod()`, that's why I wanted it external in the first place..
but maybe I can't just create another there. will try

phoracek: You can spawn a new client in each helper function. We commonly do it in KubeVirt e2e tests. It is some overhead, but it keeps the code cleaner to read, which should be the priority on tests.

RamLavi: I preferred to pass the client.. let me know what you think

tests/ovs_test.go
RamLavi: Done

tests/ovs_test.go
RamLavi: Done

(but to understand, can you explain why?)

tests/ovs_test.go
RamLavi: Done

tests/ovs_test.go
RamLavi: Done

tests/ovs_test.go
RamLavi: ack. 
Done

tests/ovs_test.go
RamLavi: Done

tests/cluster/cluster.go
RamLavi: Moved to interface, initiated in tests_suite_test.go

RamLavi:
@phoracek [fixed #2](https://github.com/kubevirt/ovs-cni/compare/811d4d0403d74c0d9a7e9ba9def1458b69390467..9fa127c9fea96de6a61f9f8aec48f9cbfe03b6fc)

tests/ovs_test.go
ormergi: You need to use `EventualyWithOffset` when you assert on helpers functions,
so when an error occurs the output will show the line number of the function caused the error on the test code file (instead of the helpers code file).
https://github.com/onsi/gomega/blob/master/gomega_dsl.go#L160

The integer passed to `EventualyWithOffset` is used to calculate the offset in the call stack, `Eventually` warps `EventualyWithOffset` with offset = 0 
https://github.com/onsi/gomega/blob/master/gomega_dsl.go#L206

For example:  you assert on helper function (level 2):
0    Test code (`It`block) 
1             AssertHelper
If an error occurs and you want to return the error line number of the test code level, you need to use `EventualWithOffset(1,..) at `AssertHelper`

As you already know we encourage developers to use assertions in test code only and not in helpers to make the tests easier to understand and maintain :)

ormergi (approved):

tests/cluster/cluster.go
```diff
@@ -0,0 +1,186 @@
...
+func (api *ClusterApi) CreatePrivilegedPodWithIp(podName, nadName, cidr string) {
+	By(fmt.Sprintf("Creating pod %s with priviliged premission and ip %s", podName, cidr))
+	privileged := true
+	podObject := &corev1.Pod{ObjectMeta: metav1.ObjectMeta{
+		Name: podName,
+		Namespace: testNamespace,
+		// This annotation makes sure the pod is assigned to a node that has this ovs bridge resource
+		Annotations: map[string]string{"k8s.v1.cni.cncf.io/networks": nadName},
+	},
+	Spec: corev1.PodSpec{ Containers: []corev1.Container{{Name: "test",
```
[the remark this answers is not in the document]
RamLavi: Done

tests/cluster/cluster.go
```diff
@@ -0,0 +1,186 @@
...
+		Clientset: clientset,
+		NetClient: netClient,
+		RestConfig: restConfig,
+	}
+}
+
+func (api *ClusterApi) CreateTestNamespace() {
+	By(fmt.Sprintf("Creating %s namespace", testNamespace))
+	_, err := api.Clientset.CoreV1().Namespaces().Create(context.TODO(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: testNamespace}}, metav1.CreateOptions{})
+	Expect(err).ToNot(HaveOccurred(), "should succeed creating test namespace")
```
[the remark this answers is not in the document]
RamLavi: Done

tests/cluster/cluster.go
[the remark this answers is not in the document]
RamLavi: Done

tests/cluster/cluster.go
```diff
@@ -0,0 +1,186 @@
...
+				if containerStatus.Ready != true {
+					return false
+				}
+			}
+
+			return true
+		}
+		return false
+
+	}, 120*time.Second, 5*time.Second).Should(BeTrue(), "should succeed getting pod container to Ready state")
```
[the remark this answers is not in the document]
RamLavi: Done

tests/cluster/cluster.go
```diff
@@ -0,0 +1,186 @@
...
+
+	for _, pod := range podList.Items {
+		err = api.Clientset.CoreV1().Pods(testNamespace).Delete(context.TODO(), pod.Name, metav1.DeleteOptions{})
+		Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("should succeed deleting pod %s", pod.Name))
+	}
+
+	Eventually(func() int {
+		podsList, err := api.Clientset.CoreV1().Pods(testNamespace).List(context.TODO(), metav1.ListOptions{})
+		Expect(err).ToNot(HaveOccurred(), "Should succeed getting pod list in test namespace after deletion")
+		return len(podsList.Items)
```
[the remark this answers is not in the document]
RamLavi: Done

tests/pod/pod.go
```diff
@@ -0,0 +1,73 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/remotecommand"
+	"github.com/pkg/errors"
+)
+
+func ExecOnPod(clientset *kubernetes.Clientset, restConfig *restclient.Config, podName, containerName, namespace, command string) (string, string, error) {
```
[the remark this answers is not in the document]
RamLavi: Done

tests/cluster/cluster.go
```diff
@@ -0,0 +1,186 @@
...
+	By("Removing ovs-bridge on the node")
+	out, err := node.RunOnNode("node01", "sudo ovs-vsctl --if-exists del-br " + bridgeName)
+	Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("failed to run command on node. stdout: %s", out))
+}
+
+func (api *ClusterApi) AddOvsBridgeOnNode(bridgeName string) {
+	By("Adding ovs-bridge on the node")
+	out, err := node.RunOnNode("node01", "sudo ovs-vsctl add-br " + bridgeName)
+	Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("failed to run command on node. stdout: %s", out))
+}
```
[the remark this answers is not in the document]
RamLavi: Done

tests/ovs_test.go
```diff
@@ -33,4 +37,51 @@ var _ = Describe("ovs-cni", func() {
...
+						clusterApi.DeletePodsInTestNamespace()
+					})
+
+					Specify("they should be able to communicate over the network", func() {
+						By("Checking pods connectivity by pinging from one to the other")
+						ipPod2, _, err := net.ParseCIDR(cidrPod2)
+						Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("should succeed parsing pod's cidr: %s", cidrPod2))
+
+						err = clusterApi.PingFromPod(pod1Name, "test", ipPod2.String())
+						Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("should be able to ping from pod %s to IP %s which is on pod %s", pod1Name, ipPod2, pod2Name))
```
ormergi, line 81: suggestion: more information on ping failure: 
`fmt.Sprintf("should be able to ping from pod '%s@%s' to pod '%s@%s'", pod1.Name, pod1.IP, pod2.Name, pod2.IP`)

ormergi (approved):
Very nice!

tests/ovs_test.go
RamLavi: [Done](https://github.com/kubevirt/ovs-cni/compare/2f89681ada6776e54cce3f2e1229bbbe20bd49c1..4c383b8b461f651e74455bf35659de11765ed1df)

phoracek (approved):
/lgtm
/approve

PR #128, opened 2020-09-23
Bump openvswitch to version 2.13

rhrazdil:
DPDK dependency of openvswitch can now be installed directly from centos 8 extras repo.
New openvswitch version outputs interface names without doublequotes, addressing this
change in the plugin_tests.go.

Signed-off-by: Radim Hrazdil <rhrazdil@redhat.com>

### DIFF
--- a/cluster/up.sh
+++ b/cluster/up.sh
@@ -23,7 +23,11 @@ $(kubevirtci::path)/cluster-up/up.sh
 
 echo 'Installing Open vSwitch on nodes'
 for node in $(./cluster/kubectl.sh get nodes --no-headers | awk '{print $1}'); do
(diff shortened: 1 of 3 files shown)


phoracek (approved):
/lgtm
/approve

Thanks!

rhrazdil:
@phoracek Now I realized that CI is running `make docker-test`, bumped fedora to 32 in Dockerfile, which has the up to date openvswitch version.

phoracek:
But that is only for our integration tests, not for e2e. And even then, it still uses the kernel module from the host. So overall it is a good idea to keep this reasonably up-to-date.

rhrazdil:
> But that is only for our integration tests, not for e2e. And even then, it still uses the kernel module from the host. So overall it is a good idea to keep this reasonably up-to-date.

Hmm, I tried printing out the ovs-vsctl version and with fedora 31 there is 2.12 and with fedora 32 there is 2.13, so it does make a difference.

@RamLavi

RamLavi:
/lgtm

great thanks!

PR #129, opened 2020-09-28
handle smartnic cleanup when container net namespace is empty

pperiyasamy:
This change attempts to cleanup smartVFs net representor from
ovs bridge and rename smartVF back to its original name when
container network namespace is no more available due to errors
like OOMKilled, CrashLoopBackOff etc.

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -446,7 +446,7 @@ func CmdDel(args *skel.CmdArgs) error {
 			}
 			if err = removeOvsPort(ovsDriver, rep); err != nil {
 				// Don't throw err as delete can be called multiple times because of error in ResetVF and ovs
-				// port is already deleted in a previous invocation. so log it and proceed further.
+				// port is already deleted in a previous invocation.
 				log.Printf("Error: %v\n", err)
 			}
 			if err = sriov.ResetVF(args, netconf.DeviceID); err != nil {
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -437,9 +437,27 @@ func CmdDel(args *skel.CmdArgs) error {
 	if args.Netns == "" {
 		// The CNI_NETNS parameter may be empty according to version 0.4.0
 		// of the CNI spec (https://github.com/containernetworking/cni/blob/spec-v0.4.0/SPEC.md).
-		// In accordance with the spec we clean up as many resources as possible.
-		if err := cleanPorts(ovsDriver); err != nil {
-			return err
+		if netconf.DeviceID != "" {
+			// SR-IOV Case - The sriov device is moved into host network namespace when args.Netns is empty.
+			// This happens container is killed due to an error (example: CrashLoopBackOff, OOMKilled)
+			var rep string
+			if rep, err = sriov.GetNetRepresentor(netconf.DeviceID); err != nil {
+				return err
+			}
+			if err = removeOvsPort(ovsDriver, rep); err != nil {
+				// Don't throw err as delete can be called multiple times because of error in ResetVF and ovs
+				// port is already deleted in a previous invocation. so log it and proceed further.
+				log.Printf("removal of ovs port %s is failed for device %s , it may be removed already"+
+					" by previous delete call, err %v", rep, netconf.DeviceID, err)
+			}
+			if err = sriov.ResetVF(args, netconf.DeviceID); err != nil {
+				return err
+			}
+		} else {
+			// In accordance with the spec we clean up as many resources as possible.
+			if err := cleanPorts(ovsDriver); err != nil {
+				return err
+			}
 		}
 		return nil
 	} else {
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -447,8 +447,7 @@ func CmdDel(args *skel.CmdArgs) error {
 			if err = removeOvsPort(ovsDriver, rep); err != nil {
 				// Don't throw err as delete can be called multiple times because of error in ResetVF and ovs
 				// port is already deleted in a previous invocation. so log it and proceed further.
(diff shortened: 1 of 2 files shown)


pkg/plugin/plugin.go
```diff
@@ -437,9 +437,27 @@ func CmdDel(args *skel.CmdArgs) error {
...
 		// The CNI_NETNS parameter may be empty according to version 0.4.0
 		// of the CNI spec (https://github.com/containernetworking/cni/blob/spec-v0.4.0/SPEC.md).
-		// In accordance with the spec we clean up as many resources as possible.
-		if err := cleanPorts(ovsDriver); err != nil {
-			return err
+		if netconf.DeviceID != "" {
+			// SR-IOV Case - The sriov device is moved into host network namespace when args.Netns is empty.
+			// This happens container is killed due to an error (example: CrashLoopBackOff, OOMKilled)
+			var rep string
+			if rep, err = sriov.GetNetRepresentor(netconf.DeviceID); err != nil {
```
rhrazdil, line 444: Please, use `:=` assignment instead of `var rep string`

pkg/sriov/sriov.go
```diff
@@ -58,6 +59,34 @@ func getVFLinkName(pciAddr string) (string, error) {
...
+	// smart VF attached inside the container by device plugin. It can be considered
+	// as one end of veth pair whereas other end is smartVF. The VF representor would
+	// get added into ovs bridge for the control plane configuration.
+	rep, err := sriovnet.GetVfRepresentor(uplink, vfIndex)
+	if err != nil {
+		return "", err
+	}
+
+	return rep, nil
+
```
rhrazdil, line 87: Extra newline

pkg/sriov/sriov.go
```diff
@@ -226,3 +239,32 @@ func ReleaseVF(args *skel.CmdArgs) error {
...
+	}
+	// get smart VF netdevice from PCI
+	vfNetdevices, err := sriovnet.GetNetDevicesFromPci(deviceID)
+	if err != nil {
+		return err
+	}
+	// Make sure we have 1 netdevice per pci address
+	if len(vfNetdevices) != 1 {
+		// This would happen if netdevice is not yet visible in default network namespace.
+		// so return ErrLinkNotFound error so that multus can attempt multiple times
```
rhrazdil, line 257: Is this behaviour specific for multus only?

pkg/plugin/plugin.go
```diff
@@ -437,9 +437,27 @@ func CmdDel(args *skel.CmdArgs) error {
...
+				return err
+			}
+			if err = removeOvsPort(ovsDriver, rep); err != nil {
+				// Don't throw err as delete can be called multiple times because of error in ResetVF and ovs
+				// port is already deleted in a previous invocation. so log it and proceed further.
```
rhrazdil, line 449: `so log it and proceed further.` seems superfluous

pkg/plugin/plugin.go
```diff
@@ -437,9 +437,27 @@ func CmdDel(args *skel.CmdArgs) error {
...
+				log.Printf("removal of ovs port %s is failed for device %s , it may be removed already"+
+					" by previous delete call, err %v", rep, netconf.DeviceID, err)
```
rhrazdil, line 451: In other places in this module, we just log.Printf("Error: %v\n", err) in this situation, it may be better to stick with that.

rhrazdil (changes requested):
Thanks, few comments

pkg/plugin/plugin.go
pperiyasamy: Couldn't use := as rep variable is used at line#447 (outside if scope).

pkg/plugin/plugin.go
pperiyasamy: this was needed so that ResetVF is called at line#452 to rename the smartvf back to original name in case ResetVF thrown ErrLinkNotFound error in previous CmdDel call (noticed this behaviour in my tests).

pkg/plugin/plugin.go
pperiyasamy: ok, good. done now.

pkg/sriov/sriov.go
pperiyasamy: done.

pkg/sriov/sriov.go
pperiyasamy: this can happen with any meta plugins that use ovs-cni, but I've tested only with multus. now changed it into generic term meta plugin.

pkg/plugin/plugin.go
rhrazdil: Ah, sorry for not being cear enough. I meant the sentence in the comment: `so log it and proceed further.`, I think it's not necessary as it's obvious from the next two lines.

pkg/plugin/plugin.go
rhrazdil: I see, thanks

pkg/sriov/sriov.go
rhrazdil: Thank you

pkg/plugin/plugin.go
pperiyasamy: ah! yes. is it ok now ?

rhrazdil: Thanks! Looks good to me

phoracek (commented):
/approve

:rocket:

phoracek:
@pperiyasamy do you need a new release to consume this?

pperiyasamy:
> do you need a new release to consume this?

yes @phoracek, it would be great if we have a release with this fix.

phoracek:
Sure thing sir! Here you go https://github.com/kubevirt/ovs-cni/releases/tag/v0.14.0

PR #130, opened 2020-09-29
update demo.md for ipam configuration

pperiyasamy:
Update `Configure IP Address` section in `demo.md` as ovs-cni
already have support for IPAM configuration.

### DIFF
--- a/docs/demo.md
+++ b/docs/demo.md
@@ -185,6 +185,10 @@ spec:
 EOF
 ```
 
+The `ovs-ipam-net` nework uses `host-local` IPAM plugin which works only on single
+worker node and should be replaced by some cluster-level IPAM across worker nodes
+in a K8s cluster.
+
 Now create a pod and connect it with `ovs-ipam-net` network. Now the pod container
 is also created with `net1` interface configured with ip address from
 `192.168.1.0/24` subnet.
--- a/docs/demo.md
+++ b/docs/demo.md
@@ -153,11 +153,41 @@ kubectl exec samplepod-1 -- ip link show
 
 ## Configure IP Address
 
-Open vSwitch CNI does not support IPAM yet. In order to test IP connectivity
-between Pods, we need to configure IP manually.
+Open vSwitch CNI has support for IPAM. In order to test IP connectivity
+between Pods, we first need to create `NetworkAttachmentDefinition` object
+with required IPAM configuration.
 
-Create second Pod, this time it will configure its own IP address. Note that in
-order to configure IP address, this Pod must be in privileged mode.
+```shell
+cat <<EOF | kubectl create -f -
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: ovs-ipam-net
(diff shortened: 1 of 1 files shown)


docs/demo.md
```diff
@@ -153,11 +153,41 @@ kubectl exec samplepod-1 -- ip link show
...
+  annotations:
+    k8s.v1.cni.cncf.io/resourceName: ovs-cni.network.kubevirt.io/br1
+spec:
+  config: '{
+      "cniVersion": "0.3.1",
+      "type": "ovs",
+      "bridge": "br1",
+      "vlan": 100,
+      "ipam": {
+        "type": "host-local",
```
phoracek, line 175: Would you mention in a note that we expect only a single worker - so they will share the IPAM and we won't  get duplicated IPs?

phoracek (changes requested):
Looks great, thanks for the update. Just a small nit

docs/demo.md
JanScheurich: @phoracek Fair enough. I'm not a friend of host-local IPAM in NAD examples for exactly this reason. Is there another widely available IPAM plugin you would suggest to use instead? If not, let's follow your advice and just add a disclaimer that host-local IPAM is an example for single worker demo scenarios.

phoracek: We could go with https://github.com/openshift/whereabouts-cni, but I feel it is little too big for a basic example.

What about using `static` and passing the IP as an attribute on the network request? https://github.com/containernetworking/plugins/tree/master/plugins/ipam/static

JanScheurich: @phoracek Using static IPAM wouldn't make the example simpler. Then I'd rather add a disclaimer comment that host-local IPAM plugin only works for single nodes and should be replaced by some cluster-level IPAM in real K8s clusters.

phoracek: Sounds good to me

pperiyasamy: @JanScheurich @phoracek Thanks for your review comment! updated the doc, is it ok now ?

JanScheurich: Sorry Peri. I should have paid attention and looked at the complete demo document. As it contains pod manifests and you must specify the IP address to be pinged anyhow, using 'static' IPAM plugin is actually better. You specify the static IPs to be used in the NW annotation and can safely use them in the ping command.

JanScheurich: Thanks Peri. I don't see how your example ping can have worked. The two pods are not on the same network and their IPs are in different subnets. This would require a router in between. Why don't you just spin up two pods on the same ovs-ipam -net with addresses in the same subnet?

pperiyasamy: @JanScheurich Here the IPs are assigned from 10.10.10.0/24 subset for the two pods connected over vlan 100 network. I hope it should be fine or do you also want to mention with details about cross connect b/w worker ovs instances ?

phoracek:
@JanScheurich are you fine with the current state?

/lgtm
/hold

phoracek:
/hold cancel

Thanks for the contribution @pperiyasamy !

Issue #133, opened 2020-10-30
OVS-CNI not working with sbr chained plugin due to interface index mismatch


gabidrg:
Hi,

When using OVS-CNI and chaining the sbr plugin in a NetworkAttachmentDefinition, sbr fails to create a default route in the pod. 

Steps to replicate:

1. Using the local cluster deployment with 1 master node and 2 worker nodes. Manually install sbr plugin in `/opt/cni/bin/` on every node.

2. OVS setup on every node:

```
ovs-vsctl add-br br-data 
ovs-vsctl add-port br-data eth1
```

3. I used a per-worker configuration for the net-attach-def in order to be able to allocate IPs with host-local IPAM and avoid collisions. Each node has a `/etc/cni/multus/net.d/signet-100.conflist` file using a different allocation range:

```
[root@node01 ~]# mkdir -p /etc/cni/multus/net.d
[root@node01 ~]# cat << EOF > /etc/cni/multus/net.d/signet-100.conflist
{
  "cniVersion": "0.3.1",
  "name": "signet-100",
  "plugins": [
    {
      "type": "ovs",
      "bridge": "br-data",
      "vlan": 100,
      "ipam": {
        "type": "host-local",
        "subnet": "192.168.33.0/24",
        "rangeStart": "192.168.33.150",
        "rangeEnd": "192.168.33.154",
        "routes": [
          { "dst": "0.0.0.0/0" }
        ]
      }
    },
    {
      "type": "sbr"
    }
  ]
}
EOF

[root@node03 ~]# mkdir -p /etc/cni/multus/net.d
[root@node03 ~]# cat << EOF > /etc/cni/multus/net.d/signet-100.conflist
{
  "cniVersion": "0.3.1",
  "name": "signet-100",
  "plugins": [
    {
      "type": "ovs",
      "bridge": "br-data",
      "vlan": 100,
      "ipam": {
        "type": "host-local",
        "subnet": "192.168.33.0/24",
        "rangeStart": "192.168.33.170",
        "rangeEnd": "192.168.33.174",
        "routes": [
          { "dst": "0.0.0.0/0" }
        ]
      }
    },
    {
      "type": "sbr"
    }
  ]
}
EOF
```
4. Added the net-attach-def:

```
$ cat <<EOF | ./cluster/kubectl.sh apply -f -
apiVersion: "k8s.cni.cncf.io/v1"
kind: NetworkAttachmentDefinition
metadata:
  name: signet-100
  annotations:
[...]

PR #134, opened 2020-10-30
Fix interface index in result.IPs

gabidrg:
Fixing GH-133. IPAM interface configuration is called with `result.Interfaces[0]` being the container interface. When IPAM configuration is finished, the `result.Interfaces` list is reordered and host interface is added at index 0 before the container interface, while `result.IPs` list of IP configuration information still points to container interface at index 0. This fixes interface index reference in IP configuration list so that chained plugins that are making use of these structures (such as sbr) will function properly.

Signed-off-by: Gabriel Dragomir <gabidrg@gmail.com>

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -367,6 +367,15 @@ func CmdAdd(args *skel.CmdArgs) error {
 		}
 		result = newResult
 		result.Interfaces = []*current.Interface{hostIface, result.Interfaces[0]}
+
+		for ifIndex, ifCfg := range result.Interfaces {
+			// Adjust interface index with new container interface index in result.Interfaces
+			if ifCfg.Name == args.IfName {
+				for ipIndex, _ := range result.IPs {
+					result.IPs[ipIndex].Interface = current.Int(ifIndex)
+				}
+			}
+		}
 	}
 
 	return types.PrintResult(result, netconf.CNIVersion)
--- a/pkg/plugin/plugin_test.go
+++ b/pkg/plugin/plugin_test.go
@@ -119,6 +119,9 @@ var _ = Describe("CNI Plugin", func() {
 
 		Expect(len(result.IPs)).To(Equal(1))
 
+		By("Checking that IP config in result of ADD command is referencing the container interface index")
+		Expect(result.IPs[0].Interface).To(Equal(current.Int(1)))
+
 		err = targetNs.Do(func(ns.NetNS) error {
 			defer GinkgoRecover()
 


pkg/plugin/plugin.go
```diff
@@ -367,6 +367,15 @@ func CmdAdd(args *skel.CmdArgs) error {
 		}
 		result = newResult
 		result.Interfaces = []*current.Interface{hostIface, result.Interfaces[0]}
+
+		for ifIndex, ifCfg := range result.Interfaces {
```
phoracek, line 371: This looks valid, but I believe we could fix the issue at its source, on line 359. Would you please investigate that option?

phoracek (changes requested):
Thanks a lot for the opened Issue and this contribution.

/ok-to-test

pkg/plugin/plugin.go
gabidrg: Hello and thanks for the suggestion! Initially I attempted to fix at that point and did not understood why `newResult` had to be initialized there, but after reading the CNI spec and following the implementation logic, I thought like 359 is actually OK for these reasons:
- according to the CNI spec the IPAM returns no interfaces key: 

> Note that unlike regular CNI plugins, IPAM plugins should return an abbreviated Result structure that does not include the interfaces key, since IPAM plugins should be unaware of interfaces configured by their parent plugin except those specifically required for IPAM (eg, like the dhcp IPAM plugin).

- for this reason `newResult.Interfaces` is formed in order to be used with `ipam.ConfigureIface` and it actually only contains the container interface. Thus, line 359 correctly points to index 0.

So I chose to correct the index after IPAM is done at the end of ADD action.

phoracek: Fair enough.

Would you please extend the IPAM test to cover this too? To make sure the issue does not reappear.

phoracek:
/cc @pperiyasamy 

In case you want to review changes on the IPAM code.

pkg/plugin/plugin.go
pperiyasamy: @phoracek Do we still have to return hostIface ? This is anyway not needed by metaplugin (example: multus) to build the pod network status. If not needed, we can remove this and it would fix the ip index problem.

phoracek: Good question. I don't know about any use for the host iface. OVS CNI only attaches to the bridge, never creates it, so I think it would be safe to drop it. It would certainly make things cleaner.

@gabidrg, would you please try what Peri proposed?

gabidrg: As far as I see the veth pair is created by OVS CNI. The CNI spec mentions that host interfaces might be returned in ADD action result and I see use cases for having the host interface in result in other plugins (e.g. bridge) for instance to be used in the CHECK action (which should be implemented sooner or later for OVS CNI as well) or for handling of host side interface configuration (IP or VLAN tag?). Also, I am not sure about whether the SR-IOV use case might need it.

phoracek: I see. Then from my side, I'm just missing the test.

pperiyasamy: AFAIK the metaplugin neither store the returned interfaces and nor pass it in the CmdCheck. 
The CmdCheck method still takes *skel.CmdArgs as argument as similar to CmdAdd, so its up to OVS CNI responsibility to find out host interface, so returning host interface in CmdAdd may not help here.
I'm not sure about meta-plugin uses bandwidth on the host interface since in this case it's just a OVS port attached with OVS bridge. is bandwidth still useful here ?

gabidrg: To make sure we have the same representation of the issue, when I mentioned the result returning the interfaces, I meant the OVS-CNI plugin and its flow implementing ADD, DEL, CHECK actions and not Multus, as the metaplugin is just delegating the network configuration to OVS-CNI. In my understanding, when ADD is executed, the configuration it gets via *skel.CmdArgs triggers either the SR-IOV flow to configure a VF, or the general case when a veth pair is setup. In either cases the OVS bridge is already configured and not managed by OVS-CNI. Yet, it is OVS-CNI that manages the host side interface added to the bridge. As I mentioned above, I had two concerns about the result not containing the hostInterface:
- first in relation with its inner workings: subsequent actions such as CHECK might need it. This can be indeed solved as the DEL action does: it retrieves the hostInterface by using a query to ovsdb. So in this case, question is whether it is more efficient to keep the hostInterface in the result or not, in order to be used at some point by some CHECK action that has yet to be implemented.
- secondly in relation with chained plugins: I reported this issue because I found that sbr is not working while working for a telco use case. I also mentioned bandwitdth as other potential chained plugin that might use hostInterface, because it has use cases in telco QoS scenarios. The idea is to allow them to function properly if they need to operate over the host interface keeping in mind that they are unaware of the CNI implementation.

pperiyasamy: ah! yes, I just noticed `prevResult` in `StdinData of CmdArgs` holds result obj which is then passed into check method. Then let's keep returning hostIface in the result.

pkg/plugin/plugin.go
```diff
@@ -367,6 +367,15 @@ func CmdAdd(args *skel.CmdArgs) error {
...
+			// Adjust interface index with new container interface index in result.Interfaces
+			if ifCfg.Name == args.IfName {
+				for ipIndex, _ := range result.IPs {
+					result.IPs[ipIndex].Interface = current.Int(ifIndex)
```
pperiyasamy, line 375: can't we just hardcode ifIndex into 1 (as we knew containerIface is 2nd element in the result.Interfaces ? 
This would avoid outer for and if blocks.

gabidrg: I could do that. I usually write quite defensive code, the outer loops are there to protect against an eventual re-ordering of interfaces in the result.

pperiyasamy: The result.Interfaces is just created locally before this block, an array and ordering is guaranteed (as per insertion of elements), we can just believe on this invariant.  But I just leave it to you to decide on the approach.

phoracek:
Thanks for the contribution. I created a new patch release for this https://github.com/kubevirt/ovs-cni/releases/tag/v0.16.2

gabidrg:
Thank you very much!

PR #137, opened 2020-11-05
send gratuitous arp upon assigning ip address

pperiyasamy:
Make ovs-cni to send gratuitous arp on the container interface so that other pods
to update its ARP cache with new mac address for the existing ip address.
This would enable ip reuse and supports seamless pod upgrade, scaling,
redeployment scenarios.


pkg/plugin/plugin.go
```diff
@@ -359,8 +361,44 @@ func CmdAdd(args *skel.CmdArgs) error {
 			ipc.Interface = current.Int(0)
 		}
 
+		// wait until OF port link state becomes up. This is needed to make
+		// gratuitous arp for args.IfName to be sent over ovs bridge
+		for i := 0; i < 5; i++ {
```
phoracek, line 366: Could you please define a helper function for this `waitLinkUp` taking the number of retries and interval? To keep the main flow of this (already too big) function clean.

pkg/plugin/plugin.go
```diff
@@ -359,8 +361,44 @@ func CmdAdd(args *skel.CmdArgs) error {
...
+				if portState == "up" {
+					break
+				}
+			}
+			if i == 4 {
+				log.Printf("The OF port %s state is not up until this time", hostIface.Name)
+			} else {
+				time.Sleep(500 * time.Millisecond)
+			}
+		}
```
phoracek, line 380: I don't feel comfortable with this code in. Issue is that it never fails, it only logs an error. And we don't have a test covering it. Due to that, this can just stop working and we will never notice.

I imagine testing this may become quite difficult. For me, it would be enough if we extended the existing IPAM test with an assertion making sure that the output never logs these issues (although it would be quite easy to get a false positive then).

Any other idea how to test it?

Or could we make the new code less forgiving and hard fail in case things don't go as expected?

phoracek (changes requested):

pkg/plugin/plugin.go
JanScheurich: I see no problem in letting the function fail hard in case the link state of the veth or VF representor port doesn't come up. But I wonder what is the approach in other forwarders in that case.

pkg/plugin/plugin.go
```diff
@@ -43,7 +43,11 @@ import (
 	"github.com/kubevirt/ovs-cni/pkg/sriov"
 )
 
-const macSetupRetries = 2
+const (
+	macSetupRetries        = 2
+	linkstateCheckRetries  = 5
+	linkStateCheckInterval = 500 // in milliseconds
```
phoracek, line 49: You can set it to 500 * time.Millisecond.

phoracek (changes requested):
One nit. The rest looks great, thanks.

phoracek (approved):
Really neat. Thanks a lot.

/lgtm
/approve
/hold

Putting on hold for @JanScheurich to take a final look if he wants to.

pkg/plugin/plugin.go
pperiyasamy: @JanScheurich the code is now changed to use fail fast approach when link state is not up until given retries/interval.  This check may not be needed in CNIs like [macvlan ](https://github.com/containernetworking/plugins/blob/master/plugins/main/macvlan/macvlan.go#L318)because that doesn't have something OF port which comes operationally up bit later after veth link is up.

pkg/plugin/plugin.go
pperiyasamy: Good thought! done.

JanScheurich:
I'm happy with the final result.

/lgtm

pperiyasamy:
> The following test **failed**, say `/retest` to rerun all failed tests:

Looks timeout error occurred while pulling openvswitch rpm, triggered the retest. 
Let's see if it works now.

phoracek:
@RamLavi could this be caused by any of the recent changes on the test infra?

https://storage.googleapis.com/kubevirt-prow/pr-logs/pull/kubevirt_ovs-cni/137/pull-e2e-ovs-cni/1326171243224764416/build-log.txt?Expires=1605023864&GoogleAccessId=prow-46%40kubevirt-240711.iam.gserviceaccount.com&Signature=fVUx8zb8ZhGmPUNxX5tx07xMZFGTM2G05LwG97zkWcwjZv6NqMyqUvLxaikr1N3xK%2FWfnHX8icEILoq%2BAEpBut3H1yYITNDVs2daHn64UxY63RrOlp0cCcAgEI7uIRhAc%2B525ZMDkTeqXC56BAkd0rVMaJvt3zm%2F5sCfqvDQBKEB8Hotc9D9kW0vqKYrzNrQ4uLNW8yu%2FZnsIPL2L7bH1Qb5qVoiLV8m6xZRHYGWcXvBjLh3vY3YKCg%2FAqOFiCrayiTd5GtycZagyYKqdGH1yT%2BH10yIu%2Bpp5hWa7h87ak3AyFFPtjjrwQuYJ2oPxY3WgCXNFU%2Fb1Yw5yLQVt57I9Q%3D%3D

PR #138, opened 2020-11-09
go-install, Add vendored go

RamLavi:
Currently tier1 tests cannot run if go
is not installed in the host. installing it
internally reduces indirections and increases
stability of code.
Added vendored go install according to version set
in go.mod

This PR fixes an [error](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_project-infra/582/rehearsal-pull-e2e-cluster-network-addons-operator-ovs-cni-functests/1325733986491699200#1:build-log.txt%3A946) running the ovs-cni tier1 lane on CNAO

### DIFF
--- a/hack/install-go.sh
+++ b/hack/install-go.sh
@@ -1,0 +1,10 @@
+#!/bin/bash -xe
+
+destination=$1
+version=$(grep "^go " go.mod | awk '{print $2}')
+tarball=go$version.linux-amd64.tar.gz
+url=https://dl.google.com/go/
+
+mkdir -p $destination
+curl -L $url/$tarball -o $destination/$tarball
+tar -xf $destination/$tarball -C $destination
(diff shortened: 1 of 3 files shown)


hack/install-go.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -xe
+
+destination=$1
+version=$(grep "^go " go.mod |awk '{print $2}')
```
rhrazdil, line 4: nit:
```suggestion
version=$(grep "^go " go.mod | awk '{print $2}')
```

rhrazdil (commented):
small formatting nit, otherwise looks good to me

hack/install-go.sh
RamLavi: [Done](https://github.com/kubevirt/ovs-cni/compare/ed5133d4b2c728f7db379512ebf579327c71582a..4968ebbd54f49e7a86c6adf9b2873bd577c7dd41)

phoracek (approved):
/approve

PR #139, opened 2020-11-11
ci, Use vendored-go in functest

RamLavi:
Currently we cannot run functests if the host 
doesn't have go installed. 
We want to be independent of this while running tests.
Added vendored-go in functest make target


phoracek (approved):
/lgtm
/approve

PR #140, opened 2020-11-11
fix golint warnings

pperiyasamy:
This fixes golint code style warnings across go modules in the project.
It is makes golint a mandatory prerequisite for CI to pass.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,15 +11,36 @@ BIN_DIR = $(CURDIR)/build/_output/bin/
 export GOROOT=$(BIN_DIR)/go/
 export GOBIN = $(GOROOT)/bin/
 export PATH := $(GOBIN):$(PATH)
+GOPATH = $(CURDIR)/.gopath
+ORG_PATH = github.com/kubevirt
+PACKAGE = ovs-cni
+REPO_PATH = $(ORG_PATH)/$(PACKAGE)
+BASE = $(GOPATH)/src/$(REPO_PATH)
+PKGS = $(or $(PKG),$(shell cd $(BASE) && env GOPATH=$(GOPATH) $(GO) list ./... | grep -v "$(PACKAGE)/vendor/" | grep -v "$(PACKAGE)/tests/cluster" | grep -v "$(PACKAGE)/tests/node"))
+V = 0
+Q = $(if $(filter 1,$V),,@)
 
-all: build
+all: lint build
 
 GO := $(GOBIN)/go
 
 $(GO):
 	hack/install-go.sh $(BIN_DIR)
 
+$(BASE): ; $(info  setting GOPATH...)
+	@mkdir -p $(dir $@)
+	@ln -sf $(CURDIR) $@
+
+GOLINT = $(GOBIN)/golint
+$(GOBIN)/golint: | $(BASE) ; $(info  building golint...)
+	$Q go get -u golang.org/x/lint/golint
+
 build: format $(patsubst %, build-%, $(COMPONENTS))
+
+lint: | $(GO) $(BASE) $(GOLINT) ; $(info  running golint...) @ ## Run golint
+	$Q cd $(BASE) && ret=0 && for pkg in $(PKGS); do \
+		test -z "$$($(GOLINT) $$pkg | tee /dev/stderr)" || ret=1 ; \
+	 done ; exit $$ret
 
 build-%: $(GO)
 	hack/version.sh > ./cmd/$(subst -,/,$*)/.version
@@ -67,4 +88,4 @@ cluster-down:
 cluster-sync: build
 	./cluster/sync.sh
 
-.PHONY: build format test docker-build docker-push dep clean-dep manifests cluster-up cluster-down cluster-sync
+.PHONY: build format test docker-build docker-push dep clean-dep manifests cluster-up cluster-down cluster-sync lint
--- a/automation/check-patch.e2e.sh
+++ b/automation/check-patch.e2e.sh
@@ -17,6 +17,9 @@ main() {
     source automation/setup.sh
     cd ${TMP_PROJECT_PATH}
 
+    echo 'Run golint'
+    make lint
+
     echo 'Run functional tests'
     make docker-test
 
--- a/pkg/plugin/plugin_test.go
+++ b/pkg/plugin/plugin_test.go
@@ -37,10 +37,10 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-const BRIDGE_NAME = "test-bridge"
-const VLAN_ID = 100
-const MTU = 1456
-const DEFAULT_MTU = 1500
+const bridgeName = "test-bridge"
+const vlanID = 100
+const mtu = 1456
+const defaultMTU = 1500
 
 var _ = BeforeSuite(func() {
 	output, err := exec.Command("ovs-vsctl", "show").CombinedOutput()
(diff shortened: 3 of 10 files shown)


phoracek:
Thanks @pperiyasamy. I'm sure we'd break this again unless we introduce a CI check. Would you please add a new make target with this sanity check and called it from https://github.com/kubevirt/ovs-cni/blob/master/automation/check-patch.e2e.sh#L19?

tests/cluster/cluster.go
```diff
@@ -41,50 +38,58 @@ import (
...
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/remotecommand"
+
+	test "github.com/onsi/ginkgo"
+	matcher "github.com/onsi/gomega"
 )
 
-type ClusterApi struct {
+// Cluster object containing k8s cni cncf api client referneces
+type Cluster struct {
```
phoracek, line 47: Why are we renaming this? Should not it become ClusterAPI instead?

tests/node/node.go
```diff
@@ -24,11 +24,11 @@ import (
 	"os/exec"
 	"strings"
 
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
+	test "github.com/onsi/ginkgo"
+	matcher "github.com/onsi/gomega"
```
phoracek, line 28: Why are we changing this? The dot import is widely practiced in Go projects I saw. And we leave it in other files here (ovs_test.go).

pkg/plugin/plugin_test.go
```diff
@@ -332,7 +328,7 @@ var _ = Describe("CNI Plugin", func() {
 				"type": "ovs",
 				"bridge": "%s",
 				"mtu": %d
-			}`, BRIDGE_NAME, MTU)
+			}`, bridgeName, MTU)
```
phoracek, line 331: if you rename constants to lower came case, we should rename MTU too. It should not be public.

automation/check-patch.e2e.sh
```diff
@@ -29,3 +32,4 @@ main() {
 }
 
 [[ "${BASH_SOURCE[0]}" == "$0" ]] && main "$@"
+
```
phoracek, line 35: This is an empty trailing line. It can be dropped.

Makefile
```diff
@@ -67,4 +88,5 @@ cluster-down:
 cluster-sync: build
 	./cluster/sync.sh
 
-.PHONY: build format test docker-build docker-push dep clean-dep manifests cluster-up cluster-down cluster-sync
+.PHONY: build format test docker-build docker-push dep clean-dep manifests cluster-up cluster-down cluster-sync lint
+
```
phoracek, line 92: This is an empty trailing line. It can be dropped.

phoracek (changes requested):

Makefile
pperiyasamy: done.

automation/check-patch.e2e.sh
pperiyasamy: done

pkg/plugin/plugin_test.go
pperiyasamy: yes, renamed the constant.

tests/cluster/cluster.go
pperiyasamy: Yes, of course, changed it into ClusterAPI.

tests/node/node.go
pperiyasamy: I've renamed the node_test.go into node.go (because it contains no tests) as there were CI failures in an earlier commit.
when go module file name is not with *_test.go, golint throws error for dot imports. 

But it looks like CI failure was due to clusterApi referencing issue in the test suite. 
now changed it back to node_test.go and cluser_test.go. Hope CI succeeds now.

phoracek:
CI failed on `flag provided but not defined: -junit-output`. It seems unrelated. @RamLavi could it be related to recent Go changes?

tests/node/node.go
```diff
@@ -24,11 +24,11 @@ import (
...
 )
 
-// TODO: Use job with a node affinity instead
+// RunOnNode Use job with a node affinity instead
```
RamLavi, line 31: I think we should keep the TODO, and add an additional function description

```
// RunOnNode performs bash shell command on node
// TODO: Use job with a node affinity instead
```

RamLavi:
> CI failed on `flag provided but not defined: -junit-output`. It seems unrelated. @RamLavi could it be related to recent Go changes?

it passes on master

tests/node/node.go
pperiyasamy: @RamLavi Thanks! I've rebased the PR, Let's see the CI run now.

pperiyasamy:
> > CI failed on `flag provided but not defined: -junit-output`. It seems unrelated. @RamLavi could it be related to recent Go changes?
> 
> it passes on master

Yes, this is what so strange to me, i could see junit-output is passed like this `--junit-output=$ARTIFACTS/junit.functest.xml`, where is ARTIFACTS variable is set ?

pperiyasamy:
> > > CI failed on `flag provided but not defined: -junit-output`. It seems unrelated. @RamLavi could it be related to recent Go changes?
> > 
> > 
> > it passes on master
> 
> Yes, this is what so strange to me, i could see junit-output is passed like this `--junit-output=$ARTIFACTS/junit.functest.xml`, where is ARTIFACTS variable set ?

I think we may have to parse junit-output arg at [here](https://github.com/kubevirt/ovs-cni/blob/master/tests/tests_suite_test.go#L54) to solve this issue.

phoracek:
The issue was introduced by the second commit in this PR. I'd recommend you split the second PR even further since it mixes makefile, comment and code changes. Then we should be in a better place figuring out which change broke the junit parsing.

pperiyasamy:
> I'd recommend you split the second PR even further since it mixes makefile, comment and code changes. Then we should be in a better place figuring out which change broke the junit parsing.

@phoracek thanks, that's a good idea, let me split into smaller commits to see which change causes this issue.

pperiyasamy:
@phoracek we now at least know what is the root cause of the problem, the go modules inside the `tests/cluster `and `test/node` throws golint warnings like below.

```
/tmp/ovs-cni/go/path/src/github.com/kubevirt/ovs-cni/.gopath/src/github.com/kubevirt/ovs-cni/tests/cluster/cluster.go:42:2: should not use dot imports
/tmp/ovs-cni/go/path/src/github.com/kubevirt/ovs-cni/.gopath/src/github.com/kubevirt/ovs-cni/tests/cluster/cluster.go:43:2: should not use dot imports
/tmp/ovs-cni/go/path/src/github.com/kubevirt/ovs-cni/.gopath/src/github.com/kubevirt/ovs-cni/tests/cluster/cluster.go:47:6: type name will be used as cluster.ClusterAPI by other packages, and that stutters; consider calling this API
/tmp/ovs-cni/go/path/src/github.com/kubevirt/ovs-cni/.gopath/src/github.com/kubevirt/ovs-cni/tests/node/node.go:27:2: should not use dot imports
/tmp/ovs-cni/go/path/src/github.com/kubevirt/ovs-cni/.gopath/src/github.com/kubevirt/ovs-cni/tests/node/node.go:28:2: should not use dot imports
Makefile:40: recipe for target 'lint' failed
make: *** [lint] Error 1
+ EXIT_VALUE=2
+ set +o xtrace
```

To fix this, I've renamed the files into *_test.go, after renaming, each test modules needed flag parsing (actually this was causing build breakage in earlier CI runs with unknow flag error). After parsing the flags, now build is [failing ](https://prow.apps.ovirt.org/view/gcs/kubevirt-prow/pr-logs/pull/kubevirt_ovs-cni/140/pull-e2e-ovs-cni/1338511009106628608) as there is no tests in these modules.

pperiyasamy:
@phoracek i've made golint check to be skipped for tests/cluster and tests/node packages and build is passing now. i think we're okay with that.

phoracek:
Sounds good. Thanks for the patience, Peri.

phoracek (approved):
/lgtm
/approve

phoracek:
/hold

Once second, why is this assert dropped: https://github.com/kubevirt/ovs-cni/pull/140/files#diff-547c2ca2c8f324d2f8cda404f94afc1d9da30e5292aa2703f68f16683365a136L122-L124

pperiyasamy:
> why is this assert dropped:

@phoracek  oops, i guess git rebase didn't happen properly, Thanks for figuring this out.
now restored this change.

phoracek (commented):
/lgtm
/hold cancel

PR #142, opened 2020-11-18
make ovs socket file path as configurable property

pperiyasamy:
The ovs-cni currently hardcodes the ovs unix socket file path to `/var/run/openvswitch/db.sock`.
But this would change across deployments. Hence introducing a global flat file configuration so that
`socket_file`  path can be specified and this is applied for all network attachment definition objects.

### DIFF
--- a/docs/cni-plugin.md
+++ b/docs/cni-plugin.md
@@ -45,6 +45,35 @@ Another example with a trunk port and jumbo frames:
 * `mtu` (integer, optional): MTU.
 * `trunk` (optional): List of VLAN ID's and/or ranges of accepted VLAN
   ID's.
+* `configuration_path` (optional): configuration file containing ovsdb
+  socket file path, etc.
+
+### Flatfile Configuation
+
+There is one option for flat file configuration:
+
+* `configuration_path`: A file path to a OVS CNI configuration file.
+
+OVS CNI will look for the configuration in these locations, in this order:
+
+* The location specified by the `configuration_path` option.
+* `/etc/kubernetes/cni/net.d/ovs.d/ovs.conf`
+* `/etc/cni/net.d/ovs.d/ovs.conf`
+
+You may specify the `configuration_path` to point to another location should it be desired.
+
+Any options added to the `ovs.conf` are overridden by configuration options that are in the
+CNI configuration (e.g. in a custom resource `NetworkAttachmentDefinition` used by Multus CNI
+or in the first file ASCII-betically in the CNI configuration directory -- which is
+`/etc/cni/net.d/` by default).
+
+The sample content of ovs.conf (in JSON format) is as follows:
+
+```json
+{
+  "socket_file": "/usr/local/var/run/openvswitch/db.sock"
+}
+```
 
 ## Manual Testing
 
--- a/docs/cni-plugin.md
+++ b/docs/cni-plugin.md
@@ -64,7 +64,7 @@ You may specify the `configuration_path` to point to another location should it 
 
 Any options added to the `ovs.conf` are overridden by configuration options that are in the
 CNI configuration (e.g. in a custom resource `NetworkAttachmentDefinition` used by Multus CNI
-or in the first file "ASCII-betically" in the CNI configuration directory -- which is
+or in the first file ASCII-betically in the CNI configuration directory -- which is
 `/etc/cni/net.d/` by default).
 
 The sample content of ovs.conf (in JSON format) is as follows:
--- a/docs/cni-plugin.md
+++ b/docs/cni-plugin.md
@@ -64,7 +64,7 @@ You may specify the `configuration_path` to point to another location should it 
 
 Any options added to the `ovs.conf` are overridden by configuration options that are in the
 CNI configuration (e.g. in a custom resource `NetworkAttachmentDefinition` used by Multus CNI
-or in the first file ASCII-betically in the CNI configuration directory -- which is
+or in the first file "ASCII-betically" in the CNI configuration directory -- which is
 `/etc/cni/net.d/` by default).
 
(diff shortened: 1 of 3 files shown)


docs/cni-plugin.md
```diff
@@ -45,6 +45,35 @@ Another example with a trunk port and jumbo frames:
...
+
+* The location specified by the `configuration_path` option.
+* `/etc/kubernetes/cni/net.d/ovs.d/ovs.conf`
+* `/etc/cni/net.d/ovs.d/ovs.conf`
+
+You may specify the `configuration_path` to point to another location should it be desired.
+
+Any options added to the `ovs.conf` are overridden by configuration options that are in the
+CNI configuration (e.g. in a custom resource `NetworkAttachmentDefinition` used by Multus CNI
+or in the first file ASCII-betically in the CNI configuration directory -- which is
```
phoracek, line 67: "ASCII-betically"

:D

pkg/plugin/plugin.go
```diff
@@ -109,9 +114,50 @@ func loadNetConf(bytes []byte) (*netConf, error) {
...
 
+func pathExists(path string) bool {
+	_, err := os.Stat(path)
+	if err == nil {
+		return true
+	}
+	if os.IsNotExist(err) {
+		return false
+	}
+	return true
```
phoracek, line 158: This happens if the path maybe exists, however, we are not able to `stat` it. Does it make sense to continue with the execution of CNI in this case? I would vote to either return the error or just panic right away.

pkg/plugin/plugin.go
```diff
@@ -109,9 +114,50 @@ func loadNetConf(bytes []byte) (*netConf, error) {
 		return nil, fmt.Errorf("failed to load netconf: %v", err)
 	}
 
+	// default config file dir paths
+	confdirs := []string{"/etc/kubernetes/cni/net.d/ovs.d/ovs.conf", "/etc/cni/net.d/ovs.d/ovs.conf"}
```
phoracek, line 118: Could we make this a constant on the top of the file? For better visibility.

pkg/plugin/plugin.go
```diff
@@ -109,9 +114,50 @@ func loadNetConf(bytes []byte) (*netConf, error) {
 		return nil, fmt.Errorf("failed to load netconf: %v", err)
 	}
 
+	// default config file dir paths
```
phoracek, line 117: I think this is redundant

pkg/plugin/plugin.go
```diff
@@ -109,9 +114,50 @@ func loadNetConf(bytes []byte) (*netConf, error) {
...
+			if err != nil {
+				return nil, fmt.Errorf("load ovs config file %s: error: %v", confpath, err)
+			}
+			if err := json.Unmarshal(jsonBytes, flatnetConf); err != nil {
+				return nil, fmt.Errorf("parse ovs config file %s: error: %v", confpath, err)
+			}
+			break
+		}
+	}
+	// merge both netconf and flatnetConf configurations
```
phoracek, line 142: Redundant

pkg/plugin/plugin.go
phoracek, line 117: Can we move this whole new block loading the file to a separate function? And then have one top level function which would merge `loadNetConf` and `loadFlatNetConf`? Would be easier to follow IMO.

phoracek (changes requested):
I like what you did there. I just have a couple of comments with concerns for readability of the new function. Overall it looks good. Thanks.

docs/cni-plugin.md
pperiyasamy: done

pkg/plugin/plugin.go
pperiyasamy: removed

pkg/plugin/plugin.go
pperiyasamy: this is a slice of string type, can't make it as a constant. just moved it into a method.

pkg/plugin/plugin.go
pperiyasamy: done

pkg/plugin/plugin.go
pperiyasamy: Yes, Correct! done.

pkg/plugin/plugin.go
pperiyasamy: Good Point! done it now.

docs/cni-plugin.md
phoracek: Oh, that was not a request for change, I just liked the way you name it :)

pkg/plugin/plugin.go
```diff
@@ -112,6 +117,61 @@ func loadNetConf(bytes []byte) (*netConf, error) {
 	return netconf, nil
 }
 
+func loadFlatNetConf(configPath string) (*netConf, error) {
+	confDirs := getOvsConfDir()
```
phoracek, line 121: These are not directories but files.

phoracek (changes requested):
Just a nit. Other than that, I like the PR. I will give chance until the end of the week for Jan to comment.

pkg/plugin/plugin.go
pperiyasamy: yes, that makes sense. changed it.

docs/cni-plugin.md
pperiyasamy: ok ;) actually this doc and implementation approach inspired from [whereabouts cni](https://github.com/openshift/whereabouts-cni#flatfile-configuration)!

phoracek: Whereabouts rocks!

PR #146, opened 2021-01-10
Fix segfault when SR-IOV or veth interface failed to create

moshe010:
This commit add the missing error check when creating SR-IOV or veth
interface.

Signed-off-by: Moshe Levi <moshele@nvidia.com>

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -391,9 +391,15 @@ func CmdAdd(args *skel.CmdArgs) error {
 	if netconf.DeviceID != "" {
 		// SR-IOV Case
 		hostIface, contIface, err = sriov.SetupSriovInterface(contNetns, args.ContainerID, args.IfName, netconf.MTU, netconf.DeviceID)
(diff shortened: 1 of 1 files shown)


rhrazdil:
Thanks
/lgtm

pkg/plugin/plugin.go
```diff
@@ -395,6 +395,9 @@ func CmdAdd(args *skel.CmdArgs) error {
 		// General Case
 		hostIface, contIface, err = setupVeth(contNetns, args.IfName, mac, netconf.MTU)
 	}
+	if err != nil {
```
phoracek, line 398: Would you move this right bellow both of the lines catching the error? Even though it would duplicate it, I find it better than having to figure out the whole if else above, plus the global err.

phoracek (changes requested):
/ok-to-test
/release-note-none

phoracek (approved):
/lgtm
/approve

Thanks @moshe010 ! Much appreciated :)

PR #150, opened 2021-03-14
ovs-cni, Move images to quay.io

oshoval:
In order to fix the rate limit of docker.io,
did the following changes:

1. Update the images which are not part of cluster-up
to be on quay.io instead on docker.io.
2. Bump k8s-1.17 in order to take the newest one,
which already have some images pre provisioned.
3. Freeze multus yaml, and point to
the quay.io which is maintained by CNAO.


oshoval:
Passed, lets test again

/test pull-e2e-ovs-cni

cluster/up.sh
```diff
@@ -33,5 +33,6 @@ for node in $(./cluster/kubectl.sh get nodes --no-headers | awk '{print $1}'); d
 done
 
 echo 'Deploying multus'
+# Taking latest multus (image is breathing and its on docker.io)
```
oshoval, line 36: Please advice, do you want me to snapshot the yaml and change the image there to an image that we upload to quay?
or leave it as is, it will be breathing
(we might need to update the job to use proxy, but this is little other story)

oshoval: @phoracek @EdDev 
as part of the effort to move images from docker.io to quay on ovs-cni (this PR),
please suggest if we want to freeze a specific multus yaml and push its image to quay,
currently it always fetch the latest yaml + latest stable version, and its hosted on docker.io
so we must either leave it as is, or freeze both yaml and a specific multus image

oshoval, line 36: As part of the effort to move images from docker.io to quay.io,
please suggest if we want to freeze multus yaml + the image it use,
currently it always take latest multus and latest stable image, which is on quay.

tests/cluster/cluster.go
```diff
@@ -101,7 +101,7 @@ func (api *ClusterAPI) CreatePrivilegedPodWithIP(podName, nadName, bridgeName, c
 		Annotations: map[string]string{"k8s.v1.cni.cncf.io/networks": nadName},
 	},
 		Spec: corev1.PodSpec{Containers: []corev1.Container{{Name: "test",
-			Image:           "alpine",
+			Image:           "quay.io/oshoval/alpine",
```
phoracek, line 104: Take official Fedora instead if there is no official Alpine/Cirros repo on Quay.

hack/docker-builder/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM fedora:32
+FROM quay.io/oshoval/fedora:32
```
phoracek, line 1: There is official Fedora org on Quay.

cluster/up.sh
phoracek: I don't understand what you suggest here. This PR is to move to Quay only, how is freezing the yaml related?

phoracek (changes requested):

cluster/up.sh
oshoval: See please
https://github.com/kubevirt/ovs-cni/pull/150/files/deb0fb2539dcd2fc9b67dacd9b4e9389a8b1b094#diff-426e7a410d92c9ee2ea4e365697ede2f5f829539cdb3b10d623d170a326d6076R37

It fetches the latest yaml of multus, and this yaml has images which are on docker.io
docker.io/nfvpe/multus:stable
so it might be needed as part of this PR

one solution is to freeze the yaml, push the image to quay,
and patch the frozen yaml to point to the new quay image.

phoracek: Now I can understand what you mean. Let's copy the YAML, keep it under this project and point to KubeVirt's clone of Multus https://quay.io/repository/kubevirt/cluster-network-addon-multus?tab=tags.

oshoval:
Addressed comments, (beside the open discussion)
Thanks

cluster/up.sh
oshoval: Sounds good for me, thanks

one point, since the latest yaml use `nfvpe/multus:stable`

I think that the easiest way to maintain what you just suggested, 
is to take the latest yaml , and push the image `nfvpe/multus:stable` to the quay you just pointed out,
instead taking the `latest` from there that exists, and trying to find the yaml that belong to it,

if its ok by you, i would need please credentials or help to push this image

phoracek: The quay is populated by CNAO automation. Please stick to the Multus version we currently use there. Let's keep this repository as stable as possible. Testing with the recent version of Multus is not a part of this PR.

oshoval:
Just split to commits

cluster/up.sh
oshoval: Correct me please if i am wrong,
i took a look on CNAO multus yaml, and its parametrized, its not the vanilla one
https://github.com/kubevirt/cluster-network-addons-operator/blob/3cc4d064a623104b13caccee879e9c0b9af1e97a/hack/components/bump-multus.sh
This file as part of the bump change it, and we need the original one that was fetched during the last bump.
In order to keep it simple to maintain, i think we should not reverse engineer the current already parameterized yaml, 
but don't touch this code on ovs-cni atm (we have a cache proxy on the ovs-cni prow lane, so we are safe).

The action item we can take, on other PR is to save also the vanilla yaml as part of the CNAO bump,
and even `docker cp` it to the image, then we can pull the image, copy it from the container image, and then apply it,
(it won't take extra time, as the deploy of the yaml does it anyhow),
alternatively, just save the raw yaml as part of the progress, so we can take it to the ovs-cni repo.

Otherwise, we will need to do a progress of reverse engineering a parameterized yaml again and again.

Please suggest what you think, thanks
Official raw latest
[1] https://raw.githubusercontent.com/intel/multus-cni/master/images/multus-daemonset.yml
Parameterized on CNAO
[2] https://github.com/kubevirt/cluster-network-addons-operator/blob/master/data/multus/001-multus.yaml

phoracek: > Correct me please if i am wrong,
> i took a look on CNAO multus yaml, and its parametrized, its not the vanilla one.
> 
> https://github.com/kubevirt/cluster-network-addons-operator/blob/3cc4d064a623104b13caccee879e9c0b9af1e97a/hack/components/bump-multus.sh

Both the [original](https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/deployments/Dockerfile) and [CNAO's](https://github.com/kubevirt/cluster-network-addons-operator/blob/3cc4d064a623104b13caccee879e9c0b9af1e97a/hack/components/bump-multus.sh#L117) Dockerfile build Multus from its sources and drop entrypoint script to the filesystem root. They should be fairly samiliar.

> This script as part of the bump change it, and we need the original one that was fetched during the last bump, (or to RE it, but i don't think its a good idea, please see below).
> In order to keep it simple to maintain, i think we should not reverse engineer the current already parameterized yaml,
> but don't touch this code on ovs-cni atm (we have a cache proxy on the ovs-cni prow lane, so we are safe).
> 
> The action item we can take, on other PR is to save also the vanilla yaml as part of the CNAO bump,
> and even `docker cp` it to the image, then we can pull the image, copy it from the container image, and then apply it,
> (it won't take extra time, as the deploy of the yaml does it anyhow),
> alternatively, just save the raw yaml as part of the progress, so we can take it to the ovs-cni repo.

CNAO does not care about the original manifest, it should not be concerned about needs of OVS CNI.

> 
> Otherwise, we will need to do a progress of reverse engineering a parameterized yaml again and again,
> structure changes, and its prone to mistakes, and time consuming, better to do it right imho.

Sorry, but I don't see the complexity you mention. The image CNAO builds is more or less the same as the one produced by Multus project. We can simply pin to a specific version of Multus kept under KubeVirt Quay and use the same yaml as we are using now (with changed IMAGE name) to use it. One image and one manifest with pinned versions we will probably not need to change whole year around.

> 
> Please suggest what you think, thanks
> Official raw latest
> [1] https://raw.githubusercontent.com/intel/multus-cni/master/images/multus-daemonset.yml
> Parameterized on CNAO
> [2] https://github.com/kubevirt/cluster-network-addons-operator/blob/master/data/multus/001-multus.yaml
> 
> > Testing with the recent version of Multus is not a part of this PR.
> 
> I understand, i just want to say please that this is already the case, i just strive to a method that would keep it maintainable and worth the effort.

Let's keep it simple.

oshoval:
Addressed comments.

cluster/up.sh
```diff
@@ -33,5 +33,8 @@ for node in $(./cluster/kubectl.sh get nodes --no-headers | awk '{print $1}'); d
 done
 
 echo 'Deploying multus'
-./cluster/kubectl.sh create -f https://raw.githubusercontent.com/intel/multus-cni/master/images/multus-daemonset.yml
-./cluster/kubectl.sh  -n kube-system wait --for=condition=ready -l name=multus pod --timeout=300s
+MULTUS_IMAGE=quay.io/kubevirt/cluster-network-addon-multus@sha256:b7487e14aa0e4f4d0b8f6a626af7d420b4cd0d8bda2fda1eb652c310526db1f8
+cp manifests/multus-daemonset.do-not-change.yml manifests/multus-daemonset.yml
+sed -i s#docker.io/nfvpe/multus:stable\$#$MULTUS_IMAGE# manifests/multus-daemonset.yml
```
phoracek, line 38: Put the replacement into quotes to prevent formatting, commenting, variables

cluster/up.sh
```diff
@@ -33,5 +33,8 @@ for node in $(./cluster/kubectl.sh get nodes --no-headers | awk '{print $1}'); d
 done
 
 echo 'Deploying multus'
-./cluster/kubectl.sh create -f https://raw.githubusercontent.com/intel/multus-cni/master/images/multus-daemonset.yml
-./cluster/kubectl.sh  -n kube-system wait --for=condition=ready -l name=multus pod --timeout=300s
+MULTUS_IMAGE=quay.io/kubevirt/cluster-network-addon-multus@sha256:b7487e14aa0e4f4d0b8f6a626af7d420b4cd0d8bda2fda1eb652c310526db1f8
+cp manifests/multus-daemonset.do-not-change.yml manifests/multus-daemonset.yml
```
phoracek, line 37: The copy should be in gitignore

phoracek (changes requested):

oshoval:
Addressed comments.

manifests/multus-daemonset.do-not-change.yml
```diff
@@ -0,0 +1,410 @@
+---
```
phoracek, line 1: The manifest directory should contain manifests used for installation of OVS CNI. This new manifest is only a test helper. Please move it to the cluster/ directory.

phoracek (changes requested):
Almost there, one last thing

oshoval:
Addressed comments

phoracek:
/lgtm
/approve

Thanks!

PR #167, opened 2021-06-21
multus, Update multus manifest

RamLavi:
the multus manifest has gone through some changes[0].
Updated the manifest in order to stay up to date.

[0] k8snetworkplumbingwg/multus-cni#627

### DIFF
--- a/cluster/multus-daemonset.do-not-change.yml
+++ b/cluster/multus-daemonset.do-not-change.yml
@@ -11,7 +11,7 @@ spec:
     singular: network-attachment-definition
     kind: NetworkAttachmentDefinition
     shortNames:
-    - net-attach-def
+      - net-attach-def
   versions:
     - name: v1
       served: true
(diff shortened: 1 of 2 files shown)


RamLavi:
/retest
looks like quay hiccup

RamLavi:
looks like `make docker-test` is failing.
locally I don't see an issue.
let's wait and see if the issue is nontransient

phoracek (changes requested):
Would you please expand the commit message? Why such a big change, where does it come from.

RamLavi:
> Would you please expand the commit message? Why such a big change, where does it come from.

updated the commit message.
The changes relevant to us are mainly the DaemonSet name `kube-multus-ds` and the image url `ghcr.io/k8snetworkplumbingwg/multus-cni:stable`.
I initially thought that the old image `docker.io/nfvpe/multus:stable` is not available, which is what triggered this PR, but it was only a temporary thing..

So we can still bump to keep things up to date, but it's not as important as I initially thought. what do you think @phoracek ?

phoracek (approved):
/lgtm
/approve

I don't mind bumping. It may make future bumps (that would be mandatory) easier.

PR #168, opened 2021-06-23
Fix a failure of plugin test

masap:
If the host-local plugin is not in the execution path, the test will fail as
follows. This PR fixes it by installing host-local plugin.

```
$ sudo make test-pkg-plugin
...
------------------------------
CNI Plugin connecting container to a bridge with specific IPAM set for container interface
  should successfully complete ADD and DEL commands
  /home/honma/git/ovs-cni/pkg/plugin/plugin_test.go:325
STEP: Creating temporary target namespace to simulate a container
STEP: Calling ADD command
2021/06/23 15:02:49 CNI ADD was called for container ID: dummy, network namespace /var/run/netns/cnitest-a058aa90-47f2-d394-fc73-f73d576db1c1, interface name eth0, configuration: {
				"cniVersion": "0.3.1",
				"name": "mynet",
				"type": "ovs",
				"bridge": "test-bridge",
				"ipam": {
					"type": "host-local",
					"ranges": [[ {"subnet": "10.1.2.0/24", "gateway": "10.1.2.1"} ]]
				}
			}

• Failure [0.312 seconds]
CNI Plugin
/home/honma/git/ovs-cni/pkg/plugin/plugin_test.go:56
  connecting container to a bridge
  /home/honma/git/ovs-cni/pkg/plugin/plugin_test.go:278
    with specific IPAM set for container interface
    /home/honma/git/ovs-cni/pkg/plugin/plugin_test.go:314
      should successfully complete ADD and DEL commands [It]
      /home/honma/git/ovs-cni/pkg/plugin/plugin_test.go:325

      Unexpected error:
          <*errors.errorString | 0xc000339710>: {
              s: "failed to set up IPAM plugin type \"host-local\": failed to find plugin \"host-local\" in path [/home/honma/git/ovs-cni/build/_output/bin//go//bin/ /usr/local/sbin /usr/local/bin /usr/sbin /usr/bin /sbin /bin /snap/bin]",
          }
          failed to set up IPAM plugin type "host-local": failed to find plugin "host-local" in path [/home/honma/git/ovs-cni/build/_output/bin//go//bin/ /usr/local/sbin /usr/local/bin /usr/sbin /usr/bin /sbin /bin /snap/bin]
      occurred

      /home/honma/git/ovs-cni/pkg/plugin/plugin_test.go:114
------------------------------
...
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ COMPONENTS = $(sort \
 BIN_DIR = $(CURDIR)/build/_output/bin/
 export GOROOT=$(BIN_DIR)/go/
 export GOBIN = $(GOROOT)/bin/
-export PATH := $(GOBIN):$(PATH)
+export PATH := $(GOBIN):$(PATH):$(BIN_DIR)
 GOPATH = $(CURDIR)/.gopath
 ORG_PATH = github.com/k8snetworkplumbingwg
 PACKAGE = ovs-cni
@@ -51,13 +51,23 @@ format: $(GO)
 	$(GO) fmt ./pkg/... ./cmd/...
 	$(GO) vet ./pkg/... ./cmd/...
 
-test: $(GO)
+build-host-local-plugin:
+	if [ ! -f $(BIN_DIR)/host-local -a `uname` = 'Linux' ]; then\
+		cd $(BIN_DIR) && \
+		git clone https://github.com/containernetworking/plugins && \
+		cd plugins && git checkout v0.9.1 && cd .. && \
+		./plugins/build_linux.sh && \
+		cp ./plugins/bin/host-local . && \
+		rm -rf plugins; \
+	fi
+
+test: $(GO) build-host-local-plugin
 	$(GO) test ./cmd/... ./pkg/... -v --ginkgo.v
 
 docker-test:
 	hack/test-dockerized.sh
 
-test-%: $(GO)
+test-%: $(GO) build-host-local-plugin
 	$(GO) test ./$(subst -,/,$*)/... -v --ginkgo.v
 
 functest: $(GO)
--- a/docs/cni-plugin.md
+++ b/docs/cni-plugin.md
@@ -137,7 +137,6 @@ ip netns del ns1
 This plugin also have Go test coverage. To run tests, Open vSwitch must be
 installed and its service running. Since those tests configure host networking,
 they must be executed by root.
-This also needs `host-local` ipam plugin to be present in one of the `PATH` directory.
 
 ```shell
 sudo --preserve-env make test-pkg-plugin


pperiyasamy:
Thanks for fixing this! we also need fix ipam test case which doesn't cleanup IP address after the assignment, because of this subsequent test runs fail. let me raise a PR for it.

masap:
@pperiyasamy OK. I fixed it.

masap:
@pperiyasamy Thanks for your fix. I also added ```sudo rm -rf /var/lib/cni/networks/mynet```. Because ```last_reserved_ip.0``` continue growing even though ```DEL``` command submitted. So it could over ```10.1.2.0/24``` range like ```10.1.3.1```. Then test fails.

Makefile
```diff
@@ -51,13 +51,24 @@ format: $(GO)
...
 
-test: $(GO)
+test-build-host-local-plugin:
+	if [ ! -f $(BIN_DIR)/host-local -a `uname` = 'Linux' ]; then\
+		cd $(BIN_DIR);\
+		git clone https://github.com/containernetworking/plugins;\
+		cd plugins; git checkout v0.9.1; cd ..;\
+		./plugins/build_linux.sh;\
+		cp ./plugins/bin/host-local .;\
+		rm -rf plugins;\
```
phoracek, line 61:
```suggestion
		cd $(BIN_DIR) && \
		git clone https://github.com/containernetworking/plugins && \
		pushd plugins && git checkout v0.9.1 && popd && \
		./plugins/build_linux.sh && \
		cp ./plugins/bin/host-local . && \
		rm -rf plugins; \
```

Makefile
```diff
@@ -51,13 +51,24 @@ format: $(GO)
...
+	fi
+	sudo rm -rf /var/lib/cni/networks/mynet
```
phoracek, line 63: We don't need this, do we?

Makefile
```diff
@@ -51,13 +51,24 @@ format: $(GO)
 	$(GO) fmt ./pkg/... ./cmd/...
 	$(GO) vet ./pkg/... ./cmd/...
 
-test: $(GO)
+test-build-host-local-plugin:
```
phoracek, line 54:
```suggestion
build-host-local-plugin:
```

Makefile
```diff
@@ -51,13 +51,24 @@ format: $(GO)
...
+	sudo rm -rf /var/lib/cni/networks/mynet
+
+test: $(GO) test-build-host-local-plugin
 	$(GO) test ./cmd/... ./pkg/... -v --ginkgo.v
 
 docker-test:
 	hack/test-dockerized.sh
 
-test-%: $(GO)
+test-%: $(GO) test-build-host-local-plugin
```
phoracek, line 71:
```suggestion
test-%: $(GO) build-host-local-plugin
```

Makefile
phoracek: Removing the prefix since otherwise it conflicts with the test-* target.

phoracek (changes requested):

Makefile
masap: Fixed other than ```pushd/popd``` because of the error ```/bin/sh: 4: pushd: not found```.

Makefile
masap: This is needed because ```/var/lib/cni/networks/mynet/last_reserved_ip.0``` has last reserved ip address.
It counts up like ```10.1.2.1```, ```10.1.2.2```, ...
And when it over the ```10.1.2.0/24``` range like 10.1.3.1, test fails because the test checks it.
https://github.com/k8snetworkplumbingwg/ovs-cni/blob/e3af30b64fce307c99d8328fbf26e0f9ba719b73/pkg/plugin/plugin_test.go#L332

Makefile
masap: Fixed

Makefile
masap: Fixed.

Makefile
phoracek: This must be a bug in our tests, no? When DEL is called, the CNI should release the IP. Have you tried running on top of https://github.com/k8snetworkplumbingwg/ovs-cni/pull/170?

masap: > Have you tried running on top of #170?

Yes. I am using rebased tree.

> This must be a bug in our tests, no? When DEL is called, the CNI should release the IP.

No. The PR #170 works. It releases the IP. If 
But ```host-local``` IPAM plugin does not reuse the IP.
You can confirm this by using ```cnitool```. https://github.com/containernetworking/cni/tree/master/cnitool

phoracek: Ouch, that sounds like a bug in host-local then ...

If this is the case, could we use `"dataDir"` as defined in https://www.cni.dev/plugins/current/ipam/host-local/, to keep the metadata under tmp and then safely remove it as part of the test suite?

masap: > could we use ```"dataDir"```

Ah, OK. I will imlement it.

phoracek:
With the plugin now being installed as a part of the suite, could you update the docs https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/docs/cni-plugin.md#go-tests?

Makefile
masap: > And when it over the ```10.1.2.0/24``` range like ```10.1.3.1```, test fails because the test checks it.

Sorry, I have overlooked the ```ranges``` parameter to the host-local. Because of the parameter, after the ```10.1.2.254``` it go back to ```10.1.2.2```.
https://github.com/k8snetworkplumbingwg/ovs-cni/blob/e3af30b64fce307c99d8328fbf26e0f9ba719b73/pkg/plugin/plugin_test.go#L328

So we do not need ```sudo rm -rf /var/lib/cni/networks/mynet```. I will remove it.
But using ```"dataDir"``` is still good idea, so I will implement it at another PR.

masap:
> With the plugin now being installed as a part of the suite, could you update the docs

Fixed also.

Makefile
masap: Fixed.

phoracek (approved):
/lgtm
/approve

Thanks!

PR #170, opened 2021-06-24
clean up assigned ip address in ipam test

pperiyasamy:
invoke cmd del so that IP address would be cleaned from IPAM data backend. it would make subsequent test runs always pass on this test case. This may not be a problem in CI because tests are always run on a fresh system.


phoracek (approved):
/lgtm
/approve

Thanks!

PR #173, opened 2021-06-29
Modify data directory of host-local IPAM plugin

masap:
The `host-local` IPAM plugin used to use `/var/lib/cni/networks/`, but has been moved under `/tmp` so as not to hurt the existing environment.


phoracek (approved):
/lgtm
/approve

Thanks

PR #176, opened 2021-07-02
Bump libovsdb to 0.5.0

masap:
Bump libovsdb to 0.5.0.
And bump Go to 1.16 because libovsdb 0.4.0 requires it.

release-note
```release-note
Bump Go to 1.16
Bump libovsdb to 0.4.0
Bump libovsdb to 0.5.0
```

Signed-off-by: Masashi Honma masashi.honma@gmail.com


Makefile
```diff
@@ -35,6 +35,8 @@ $(BASE): ; $(info  setting GOPATH...)
 GOLINT = $(GOBIN)/golint
 $(GOBIN)/golint: | $(BASE) ; $(info  building golint...)
 	$Q go get -u golang.org/x/lint/golint
+	# golint installation modifies go.mod and it causes golint failure so run mod tidy here
```
phoracek, line 38: Can't this be avoided by calling `make dep`?

pkg/ovsdb/ovsdb.go
```diff
@@ -30,26 +32,85 @@ const (
...
 
-// Bridge Model of the Bridge table
+type (
+	// BridgeFailMode defines failure mode
+	BridgeFailMode = string
+	// BridgeProtocols defines OpenFlow Protocol version
+	BridgeProtocols = string
+)
+
+const (
```
phoracek, line 42: Please keep only values that are used in the project

pkg/ovsdb/ovsdb.go
```diff
@@ -30,26 +32,85 @@ const (
...
+
+// Bridge defines an object in Bridge table
 type Bridge struct {
-	UUID        string            `ovs:"_uuid"`
-	Name        string            `ovs:"name"`
-	OtherConfig map[string]string `ovs:"other_config"`
-	ExternalIds map[string]string `ovs:"external_ids"`
-	Ports       []string          `ovs:"ports"`
-	Status      map[string]string `ovs:"status"`
+	UUID                string            `ovsdb:"_uuid"`
```
phoracek, line 63: We don't  use most of these. Could we clean it up, so we don't bloat the module too much? This project has the benefit of not being a library but a focused application. We should only keep code that is needed for the project.

phoracek (changes requested):

Makefile
masap: No. Running ```make dep``` before ```make lint``` does not change the result.

pkg/ovsdb/ovsdb.go
masap: Ok, fixed.

pkg/ovsdb/ovsdb.go
masap: Ok, fixed.

phoracek (approved):
I'm surprised we don't need any attributes in the schema, but it seems to work anyway.

Thanks

/lgtm
/approve

PR #177, opened 2021-07-05
resources, Update minimum resources request

oshoval:
Update ovs-cni resources requests according to empiric use cases.

See https://bugzilla.redhat.com/show_bug.cgi?id=1935218

### DIFF
--- a/examples/ovs-cni.yml
+++ b/examples/ovs-cni.yml
@@ -33,6 +33,10 @@ spec:
         imagePullPolicy: IfNotPresent
         securityContext:
           privileged: true
+        resources:
+          requests:
+            cpu: "10m"
+            memory: "15Mi"
         volumeMounts:
         - name: cnibin
           mountPath: /host/opt/cni/bin
@@ -51,6 +55,10 @@ spec:
         volumeMounts:
           - name: ovs-var-run
             mountPath: /host/var/run/openvswitch
+        resources:
+          requests:
+            cpu: "10m"
+            memory: "10Mi"
         env:
           - name: NODE_NAME
             valueFrom:
--- a/manifests/ovs-cni.yml.in
+++ b/manifests/ovs-cni.yml.in
@@ -33,6 +33,10 @@ spec:
         imagePullPolicy: ${OVS_CNI_PLUGIN_IMAGE_PULL_POLICY}
         securityContext:
           privileged: true
+        resources:
+          requests:
+            cpu: "10m"
+            memory: "15Mi"
         volumeMounts:
         - name: cnibin
           mountPath: /host${CNI_MOUNT_PATH}
@@ -51,6 +55,10 @@ spec:
         volumeMounts:
           - name: ovs-var-run
             mountPath: /host/var/run/openvswitch
+        resources:
+          requests:
+            cpu: "10m"
+            memory: "10Mi"
         env:
           - name: NODE_NAME
             valueFrom:


oshoval:
@RamLavi 
need ok-to-test please

oshoval:
/release-note-none

lane passed

examples/ovs-cni.yml
```diff
@@ -51,6 +51,10 @@ spec:
         volumeMounts:
           - name: ovs-var-run
             mountPath: /host/var/run/openvswitch
+        resources:
+          requests:
+            cpu: "5m"
```
phoracek, line 56: This does not look right. With 5m CPU, it must be way too slow to report anything. Have you tried setting the limit to 5m to see how it behaves on the minimum?

phoracek (changes requested):
Don't we need this for the init container too?

examples/ovs-cni.yml
oshoval: Actual measured numbers

```
create / delete 1000 ovs-bridges |  6m / 11Mi
Suggested |  5m / 10Mi
```

And its the minimum not the maximum
If we set it to 5m it will just simulate a really really corner case where it just get what it guarrnteed
but it can try

phoracek: For the record, copying from another thread:

I know it is just a request. But it is not more useful than 0m if it does not guarantee the application being useful, and that would defy the purpose of this effort.

Setting the request value as limit is a simple way to confirm that the application will be able to serve even under the tight resource situation.

oshoval:
> Don't we need this for the init container too?

There is a bug with metric-server to measure when there are initContainers, 
But I can put the same numbers of the pod `5m / 10Mi` which is logically fine because the pod wont schedule if there are those at least, however it feels redundant for this cause, wdyt?

phoracek:
> > Don't we need this for the init container too?
> 
> There is a bug with metric-server to measure when there are initContainers,
> But I can put the same numbers of the pod `5m / 10Mi` which is logically fine because the pod wont schedule if there are those at least, however it feels redundant for this cause, wdyt?

The requirement for the init container is static, independent of cluster size, it always just copies the binary. Could you try to figure out what is the minimum needed by experimenting with limits?

examples/ovs-cni.yml
oshoval: Well with 6m we managed to server 1000 bridges, this is where the numbers came from, but np i will try

oshoval: limited to 5m CPU, 
create  / delete of 1000 ovs bridges reflected on the node yaml (and even pretty fast comparing to when the command finished)

phoracek: Awesome, sounds good then

oshoval: let me please more accurate with the comment (wanted to EDIT but wasn't fast enough)

It does take several seconds afterwards
(the original command is long, and then after it finishes it takes lets say roughly 20 - 40 seconds to finish)

phoracek: Not sure I understand. Once all the bridges are created, it takes 20 - 40 seconds to report them as resources? Did you count in the polling interval? How fast is it without the limit?

oshoval:
about initContainer:

1. limit less than 5m / 10Mi (which is what the regular container uses) - doesn't load (Init:RunContainerError)
```
  Warning  Failed     86s                kubelet            Error: failed to start container "ovs-cni-plugin": Error response from daemon: OCI runtime create failed: container_linux.go:370: starting container process caused: process_linux.go:459: container init caused: read init-p: connection reset by peer: unknown
```

2. limit exact as the regular container 5m / 10Mi
Sometimes fails on `Init:RunContainerError` but eventually managed to start

3. limit 10m / 15Mi seems ok.
tested also with just limit 10m another time.

What bothers me, is that i am not sure we really won't to mess with those numbers, because when the transient passes, 
the minimums will be already according the real container.

what do you prefer please ?
1. To not put requests on initContainer
2. to try seveal times to see 10m CPU limit is good enough and then request it on the initContainer ?
   if so we might want to also put it on the real container, but not a must,
   just good for keeping it simple, and safer for both  containers.

examples/ovs-cni.yml
oshoval: 20 - 40 second is just the extra time, until it finish report them all
lets say that you create 1000 bridges, 
when the command itself finished and the 1000 bridges are there, 800 are already reported,
just the 200 left takes more time after the original command finished

didn't count in the polling interval nor have the info without the limit.

if we do care for it still please let me know 
I think we are safe enough, just wanted to be more accurate

phoracek:
If we need 10m for the init container, but the only thing we set is 5m on the regular container, this Pod will be never able to start on a fully utilized host, no?

Can we then set 10m on the regular container and leave init container empty if you believe setting it is problematic?

oshoval:
> If we need 10m for the init container, but the only thing we set is 5m on the regular container, this Pod will be never able to start on a fully utilized host, no?

IIUC on a full utilized host setting initContainer to 10m is not our savior, because as we see in scenario 2
give it a chance with less and it will recover, and if we strict it to 10m minimum that host won't even give it a chance.

> Can we then set 10m on the regular container and leave init container empty if you believe setting it is problematic?

I will set it to 10m on both, its not problematic that bad, just maybe redundant,
10m / 10Mi on both will take us to more safety and its not that much.

Thanks

phoracek:
Let's not depend on the recovery, we should ask for enough resource to run comfortably. Let's do 10/10?

oshoval:
> Let's not depend on the recovery, we should ask for enough resource to run comfortably. Let's do 10/10?

yep on both, on it

oshoval:
btw since multus and linux bridge are also a binary copiers, and also have 5m i will change them as well to 10m/10Mi
at least right ?

oshoval:
Added requests to initContainer, and aligned the regular container to 10m/10Mi

oshoval:
about the failure
```
+ ./cluster/kubectl.sh -n kube-system wait --for=condition=ready -l name=multus pod --timeout=300s
error: no matching resources found
```
multus not related to this PR

oshoval:
changed, managed to break it with 10Mi memory limit so extended the initContainer to 15Mi

oshoval:
Thanks
please remove the hold when you are fine with it

PR #178, opened 2021-07-07
invoke ipam delete before net namespace check

pperiyasamy:
moving `ipam.ExecDel `before `args.Netns == "" `check, this makes sure ip is cleaned up in pod delete retry scenario, because some cases args.Netns is passed with empty string.

found this issue while testing whereabouts [PR](https://github.com/k8snetworkplumbingwg/whereabouts/pull/119)


phoracek (approved):
/lgtm
/approve

PR #179, opened 2021-07-08
Fix unexpectedly disappeared document

masap:
Previouly it was showed like this.

* unix:
* tcp::
* ssl::


masap:
Like this ...
prev

PR #180, opened 2021-07-08
complete del action when resources are missing

pperiyasamy:
make del function to comply with CNI spec and not throw an error when network namespace and/or ip link is missing.

This scenario is seen in a high scaled cluster during undeployment of pods.

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -559,6 +559,15 @@ func CmdDel(args *skel.CmdArgs) error {
 			}
 			return err
 		})
+		// do the following as per cni spec (i.e. Plugins should generally complete a DEL action
+		// without error even if some resources are missing)
+		if _, ok := err.(ns.NSPathNotExistErr); ok || err == ip.ErrLinkNotFound {
+			if portFound {
+				ip.DelLinkByName(portName)
+			}
+			cleanPorts(ovsDriver)
+			return nil
+		}
 	}
 
 	return err
--- a/pkg/plugin/plugin_test.go
+++ b/pkg/plugin/plugin_test.go
@@ -311,13 +311,12 @@ var _ = Describe("CNI Plugin", func() {
 				"vlan": %d
 			}`, bridgeName, vlanID)
 			It("should successfully complete ADD and DEL commands", func() {
-				targetNs, err := testutils.NewNS()
-				Expect(err).NotTo(HaveOccurred())
+				targetNs := newNS()
 				defer func() {
-					Expect(targetNs.Close()).To(Succeed())
-					Expect(testutils.UnmountNS(targetNs)).To(Succeed())
+					closeNS(targetNs)
 				}()
-				testDel(conf, testAdd(conf, true, false, "", targetNs), targetNs, true)
+				hostIfName := testAdd(conf, true, false, "", targetNs)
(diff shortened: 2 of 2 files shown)


pkg/plugin/plugin.go
```diff
@@ -559,6 +559,15 @@ func CmdDel(args *skel.CmdArgs) error {
 			}
 			return err
 		})
+		// do the following as per cni spec (i.e. Plugins should generally complete a DEL action
+		// without error even if some resources are missing)
+		if _, ok := err.(ns.NSPathNotExistErr); ok || err == ip.ErrLinkNotFound {
+			if portFound {
```
phoracek, line 565: Correct me if I'm wrong, but I believe this is already done on line 541. Is there any reason to call it again?

pperiyasamy: line 541 deletes only interface only from ovs bridge, but interface would still present. so this would try to delete veth pair when link deletion fails on container net namespace.

phoracek: Oh right, thanks

pkg/plugin/plugin.go
```diff
@@ -559,6 +559,15 @@ func CmdDel(args *skel.CmdArgs) error {
 			}
 			return err
 		})
+		// do the following as per cni spec (i.e. Plugins should generally complete a DEL action
+		// without error even if some resources are missing)
+		if _, ok := err.(ns.NSPathNotExistErr); ok || err == ip.ErrLinkNotFound {
```
phoracek, line 564: Could you please add a test for these two cases? Use the CNI to ADD, then manually break it and finally call DEL.

The logic of the cleanup code has grown substantially and we should make sure it is covered with functional tests, to avoid regressions.

phoracek (changes requested):

pkg/plugin/plugin.go
pperiyasamy: added tests covering these two cases. pls have a look.

phoracek (changes requested):
Minor nits, but looks good overall

pkg/plugin/plugin_test.go
```diff
@@ -309,7 +311,13 @@ var _ = Describe("CNI Plugin", func() {
 				"vlan": %d
 			}`, bridgeName, vlanID)
 			It("should successfully complete ADD and DEL commands", func() {
-				testAddDel(conf, true, false, "")
+				targetNs, err := testutils.NewNS()
```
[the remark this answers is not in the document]
pperiyasamy: done

pkg/plugin/plugin_test.go
```diff
@@ -309,7 +311,13 @@ var _ = Describe("CNI Plugin", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+				defer func() {
+					Expect(targetNs.Close()).To(Succeed())
+					Expect(testutils.UnmountNS(targetNs)).To(Succeed())
+				}()
+				testDel(conf, testAdd(conf, true, false, "", targetNs), targetNs, true)
```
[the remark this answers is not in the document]
pperiyasamy: done

phoracek (approved):
Awesome, thanks a lot!

/lgtm
/approve

PR #186, opened 2021-08-02
Reduce net consumption

alonSadan:
This PR tries to reduce the net consumption done by the ovs-bridge-marker. This PR introduces the same changes done in [CNI-bridge](https://github.com/kubevirt/bridge-marker/pull/35): 
- increasing the update interval
- use cache to keep track on the available bridges
- use a jitter to avoid network peeks.

### DIFF
--- a/cmd/marker/main.go
+++ b/cmd/marker/main.go
@@ -85,8 +85,10 @@ func main() {
 	}
 
 	markerCache := cache.Cache{}
-	wait.PollImmediateInfinite(time.Duration(*updateInterval)*time.Second, func() (bool, error) {
-		if time.Now().Sub(markerCache.LastRefreshTime()) >= time.Duration(*reconcileInterval)*time.Minute {
+	wait.JitterUntil(func() {
+		jitteredReconcileInterval := wait.Jitter(time.Duration(*reconcileInterval)*time.Minute, 1.2)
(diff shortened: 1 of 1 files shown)


cmd/marker/main.go
```diff
@@ -74,11 +84,27 @@ func main() {
...
+		}
+
+		err := markerApp.Update(&markerCache)
 		if err != nil {
 			glog.Fatalf("Update failed: %v", err)
 		}
-		time.Sleep(time.Duration(*pollInterval) * time.Second)
-	}
+
+	}, time.Duration(*updateInterval)*time.Second, 1.2, true, wait.NeverStop)
```
phoracek, line 109: Any chance you did not run `go fmt` on this (surprised about different `*` formatting)?

phoracek (commented):
Minor nit. Overall looks good

cmd/marker/main.go
alonSadan: Actually I did. I also re-checked and it stays in tact. Would you like me to change that? (this kind of formatting might also be needed in [marker_test](https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/tests/marker_test.go#L64) )

phoracek: If both are compatible, it's all good. Thanks for checking

alonSadan:
> /lgtm
> /approve

Thank you Petr

PR #187, opened 2021-08-05
Set max unavailable to 10%

phoracek:
The default value of maxUnavailable is 1. That serializes upgrades and
may drag for very long time. This change turns it into 10% to speed up
the upgrade process while keeping services reasonably available.

https://bugzilla.redhat.com/show_bug.cgi?id=1990065


rhrazdil (approved):
/lgtm

PR #189, opened 2021-08-10
tests, marker: increase timeout

alonSadan:
The current timeout in the tests is 2 minutes. The update interval is 1 minute,
and it has a jitter of 1.2.
IF the test assertion start right after the start of the interval, 1 minute has passed ,
and for some reason the update fails, there could be more than two minutes between updates,
and the tests could fail.

This commit increases the tests timeout to 3 minutes, which should be enough.

Signed-off-by: alonsadan <asadan@redhat.com>

### DIFF
--- a/tests/marker_test.go
+++ b/tests/marker_test.go
@@ -48,7 +48,7 @@ var _ = Describe("ovs-cni-marker", func() {
 					return false
 				}
 				return true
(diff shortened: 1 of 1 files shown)


alonSadan:
@phoracek  would you take a look?

phoracek:
Thanks!

/lgtm
/approve

PR #195, opened 2021-08-23
Bump protobuf v1.3.2

RamLavi:


RamLavi:
running the test manually, I get this error:
not sure how to tackle this, as tls.Dialer is defined by `crypto/tls` library..
```
# github.com/ovn-org/libovsdb/client
.gopath/pkg/mod/github.com/ovn-org/libovsdb@v0.6.1-0.20210820090057-e418ed547c02/client/client.go:132:14: undefined: tls.Dialer
```

phoracek:
@RamLavi could there be a mismatch between the Golang versions that we and the dependency use? But it is hard to believe Golang would change name of something in the core library

RamLavi:
> @RamLavi could there be a mismatch between the Golang versions that we and the dependency use? But it is hard to believe Golang would change name of something in the core library

it does say `note: module requires Go 1.16`, but we are already [set](https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/go.mod#L52) to `go 1.16`
also tried to `go get github.com/ovn-org/libovsdb` but it didn't help..

PR #204, opened 2021-09-15
Fix make cluster-sync failure with latest kubevirtci

masap:
```make cluster-sync``` fails with continuous deletion confirmation process.
```
++ ./cluster/kubectl.sh get --ignore-not-found -f ./examples/ovs-cni.yml
++ wc -l
+ [[ 1 -eq 0 ]]
+ sleep 1
++ ./cluster/kubectl.sh get --ignore-not-found -f ./examples/ovs-cni.yml
++ wc -l
+ [[ 1 -eq 0 ]]
+ sleep 1
...
```
This process expects ```./cluster/kubectl.sh get --ignore-not-found -f ./examples/ovs-cni.yml``` returns no line. But currently it returns ```"selecting docker as container runtime"```. So suppress this message by specifying container runtime with
```KUBEVIRTCI_RUNTIME```.

### CODE REVIEW

**Comment by phoracek** [comment]
Odd, I have tried this locally and the output it still `""` without the message about Docker/Podman.

/ok-to-test

**Comment by masap** [comment]
@phoracek On the CI result, it appears to be returning the expected value.
```
+ expecting_message='selecting docker as container runtime'
++ ./cluster/kubectl.sh get --ignore-not-found -f ./examples/ovs-cni.yml
+ [[ selecting docker as container runtime = selecting docker as container runtime ]]
++ ./cluster/kubectl.sh get --ignore-not-found ds ovs-cni-plugin-amd64
+ [[ selecting docker as container runtime = selecting docker as container runtime ]]
++ ./cluster/kubectl.sh get --ignore-not-found ds ovs-vsctl-amd64
+ [[ selecting docker as container runtime = selecting docker as container runtime ]]
```

phoracek:
My bad, I have not deleted the previous _kubevirtci directory. Now I can reproduce it. Waiting for this specific message seems little awkward though. Would it make sense to instead perform `wc` only on stdout? Currently we do it across both stdout and stderr, the docker messages is I believe taken from stderr (https://github.com/kubevirt/kubevirtci/blob/1ad0cbd6f8621e0d8583c921581960a46155098b/cluster-up/cluster/ephemeral-provider-common.sh#L31). What do you think?

masap:
@phoracek I do not like to watch only stdout because I would like to ensure the stderr does not show error message other than that...

The message ```selecting docker as container runtime``` is shown by container runtime auto detection functionality.
I fixed this PR to avoid the message by specifying container runtime with ```KUBEVIRTCI_RUNTIME```.

phoracek:
Then maybe we could check that the return code is 0 and stdout is empty?

My worry with the proposed approach is that it favors docker and does not leverage the container runtime auto-detection mechanism: https://github.com/kubevirt/kubevirtci/blob/1ad0cbd6f8621e0d8583c921581960a46155098b/cluster-up/cluster/ephemeral-provider-common.sh#L24, and so it may be a regression for people using Podman.

masap:
Ok, fixed.

phoracek (approved):
Looks great, kudos

/lgtm
/approve

PR #205, opened 2021-09-15
derive container mac address from ip address

pperiyasamy:
attempts to fix mac duplicates (because randomness of 3 suffix bytes not sufficient) in a big K8 cluster with many pods having secondary networks via ovs-cni.

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -21,6 +21,7 @@ package plugin
 
 import (
 	"crypto/rand"
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"log"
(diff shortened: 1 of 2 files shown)


pkg/plugin/plugin.go
```diff
@@ -366,6 +383,19 @@ func CmdAdd(args *skel.CmdArgs) error {
 		}
 
 		err = contNetns.Do(func(_ ns.NetNS) error {
+			for _, ipc := range newResult.IPs {
+				containerMac := IPAddrToHWAddr(ipc.Address.IP)
+				containerLink, err := netlink.LinkByName(args.IfName)
+				if err != nil {
+					return fmt.Errorf("failed to lookup container interface %q: %v", args.IfName, err)
+				}
+				err = assignMacToLink(containerLink, containerMac, args.IfName)
```
phoracek, line 392: What if the user passes a MAC explicitly through `requestedMac` https://github.com/k8snetworkplumbingwg/ovs-cni/pull/205/files#diff-fc137d62147eb6a9c89c0566479ff219320ca691f2505631ebacbe513bcc8108R136, this would override it, right? I think we should use the IP based MAC only as a fallback.

pperiyasamy: @phoracek done.

pkg/plugin/plugin.go
```diff
@@ -366,6 +383,21 @@ func CmdAdd(args *skel.CmdArgs) error {
 		}
 
 		err = contNetns.Do(func(_ ns.NetNS) error {
+			if mac == "" && !sriov.IsOvsHardwareOffloadEnabled(netconf.DeviceID) {
+				for _, ipc := range newResult.IPs {
```
phoracek, line 387: What is the point of cycle here? We always either fail on the first one or use the first one. No? Would it be clearer if you added an extra condition `len(newResult.IPs) >= 1` and then accessed the first IP directly? So it is clear that we care about the first IP if there is any, not that we are iterating.

phoracek (commented):
Just a minor comment, the rest looks fine

pkg/plugin/plugin.go
pperiyasamy: oh! yes. good find :) I thought `newResult.IPs` was a map and added for loop with break. removed it now.

phoracek:
This looks great, thanks Peri!

/lgtm
/approve

PR #210, opened 2021-09-30
Bump code base to CNI 1.0.0

masap:
release-note
```release-note
Bump code base to CNI 1.0.0
```


pkg/plugin/plugin.go
```diff
@@ -404,7 +403,8 @@ func CmdAdd(args *skel.CmdArgs) error {
 				return fmt.Errorf("failed to look up %q: %v", args.IfName, err)
 			}
 			for _, ipc := range newResult.IPs {
-				if ipc.Version == "4" {
+				// if ip address version is 4
+				if ipc.Address.IP.To4() != nil {
```
phoracek, line 407: Please comment in the commit message why this had to change with bump of CNI.

pkg/plugin/plugin.go
```diff
@@ -122,7 +122,16 @@ func setupVeth(contNetns ns.NetNS, contIfaceName string, requestedMac string, mt
...
+		var containerMac net.HardwareAddr
+		if requestedMac != "" {
+			mac, err := net.ParseMAC(requestedMac)
+			if err != nil {
+				return fmt.Errorf("failed to parse requested MAC  %q: %v", requestedMac, err)
+			}
+			containerMac = mac
+		}
+
+		hostVeth, containerVeth, err := ip.SetupVeth(contIfaceName, mtu, requestedMac, hostNetns)
```
phoracek, line 134: Please comment in the commit message how is this related to bump of CNI (if it is). If it is unrelated and needed, please open a separate PR for it

phoracek (changes requested):

pkg/plugin/plugin.go
masap: Ok, fixed.

pkg/plugin/plugin.go
masap: Ok, fixed. This is related to the bump of CNI.

masap: Sorry for the lack of documentation. Indeed ```SetupVeth``` modification is hard to understand. If you have any other questions, feel free to ask it !

phoracek: Thanks a lot @masap o/

pkg/plugin/plugin.go
```diff
@@ -122,7 +122,16 @@ func setupVeth(contNetns ns.NetNS, contIfaceName string, requestedMac string, mt
 	// this we will make sure that both ends of the veth pair will be removed
 	// when the container is gone.
 	err := contNetns.Do(func(hostNetns ns.NetNS) error {
-		hostVeth, containerVeth, err := ip.SetupVeth(contIfaceName, mtu, hostNetns)
+		var containerMac net.HardwareAddr
```
phoracek, line 125: I think I understand the split now. I was going to suggest we configure the "random" MAC through `SetupVeth` too, but that would not allow us to deal with a conflict... With your change, this function becomes even harder to read (to be fair, it was not great even before you posted this :). 

Could you please move this block and the one where we assign random IP to two isolated functions? It would make this code block more readable and it would also help us decouple the shared variables we have here.

I won't fight for this as your PR seems valid, so feel free to reject my request, but please consider it.

masap: Removed the ParseMAC because it is already checked in ```ip.SetupVeth()```.
https://github.com/containernetworking/plugins/blob/2a9114d1af50e23ccb020f585d2c38d7e1fc36f2/pkg/ip/link_linux.go#L46-L50

And removed MAC address assignment because ```ip.SetupVeth()``` assigns MAC addresswhen it is not specified.

pkg/plugin/plugin.go
```diff
@@ -122,7 +122,7 @@ func setupVeth(contNetns ns.NetNS, contIfaceName string, requestedMac string, mt
 	// this we will make sure that both ends of the veth pair will be removed
 	// when the container is gone.
 	err := contNetns.Do(func(hostNetns ns.NetNS) error {
-		hostVeth, containerVeth, err := ip.SetupVeth(contIfaceName, mtu, hostNetns)
+		hostVeth, containerVeth, err := ip.SetupVeth(contIfaceName, mtu, requestedMac, hostNetns)
```
phoracek, line 125: This already returns the current state of the container, right? I think we can replace the `containerLink` lookup with it.

phoracek (changes requested):

phoracek:
@pperiyasamy would you get a chance to review this one? It looks in general correct to me, but it changes the way we generate MACs for veths (instead of doing it ourselves, we leave it to the OS IIUIC). Since you invested into making IPs unique, I though you may want to take a look at this too.

pkg/plugin/plugin.go
phoracek: Great catch!

pkg/plugin/plugin.go
masap: Ah, nice ! Fixed it.

PR #211, opened 2021-09-30
extend mac randomness to 4 bytes for no ip case

pperiyasamy:
This would increase chance of mac duplicate not to occur on the interface without ip config in a big deployment

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -89,8 +89,8 @@ func getHardwareAddr(ifName string) string {
 }
 
 func generateRandomMac() net.HardwareAddr {
-	prefix := []byte{0x02, 0x00, 0x00} // local unicast prefix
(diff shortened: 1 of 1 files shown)


pkg/plugin/plugin.go
```diff
@@ -89,8 +89,8 @@ func getHardwareAddr(ifName string) string {
 }
 
 func generateRandomMac() net.HardwareAddr {
-	prefix := []byte{0x02, 0x00, 0x00} // local unicast prefix
-	suffix := make([]byte, 3)
+	prefix := []byte{0x0A, 0x58} // use private MAC prefix 0A:58
```
phoracek, line 92: Good idea not to use the obvious!

phoracek (approved):
/lgtm
/approve

pperiyasamy:
Thanks @phoracek .

PR #213, opened 2021-10-20
Add nodeSelector for Linux os

kobi86:
In order to make sure that ovs-cni can be used
in mixed node environment I added Linux OS nodeSelector


phoracek:
@kobi86 thanks for the contribution. Please post this again with the DCO signature (commit with `-s`)

PR #218, opened 2022-02-02
[release-0.19] go: Bump to 1.16

qinqon:
There is some proxy url not working with go 1.13:

parse https://proxy.golang.org|https://goproxy.io|direct: invalid character "|" in host name

This change bumps go to 1.16

It's needed to fix https://github.com/kubevirt/cluster-network-addons-operator/pull/1166


Makefile
```diff
@@ -11,6 +11,7 @@ BIN_DIR = $(CURDIR)/build/_output/bin/
 export GOROOT=$(BIN_DIR)/go/
 export GOBIN = $(GOROOT)/bin/
 export PATH := $(GOBIN):$(PATH)
+export GOFLAGS := -mod=vendor
```
rhrazdil, line 14: What does this do?

rhrazdil: Ah, nevermind, found it

automation/check-patch.e2e.sh
```diff
@@ -18,7 +18,6 @@ main() {
     cd ${TMP_PROJECT_PATH}
 
     echo 'Run golint'
```
rhrazdil, line 20: We should remove this message as well, or run run a different linter, like go vet instead?

qinqon: Let's just run `make format`

automation/check-patch.e2e.sh
```diff
@@ -17,8 +17,8 @@ main() {
     source automation/setup.sh
     cd ${TMP_PROJECT_PATH}
 
-    echo 'Run golint'
-    make lint
+    echo 'Run format'
+    make format
```
phoracek, line 21: This target does not check anything though, does it? The original target would fail the check if the code is not formatted properly, this one would just silently pass IIUIC

phoracek (changes requested):

automation/check-patch.e2e.sh
qinqon: Added `make check`

phoracek (commented):
/lgtm
/approve

PR #226, opened 2022-03-03
Remove restriction of `protobuf` version

masap:
This was introduced in #195, but `ovs-cni` is currently working without it.


phoracek:
protobuf >= 1.3.2 is needed to deal with https://nvd.nist.gov/vuln/detail/CVE-2021-3121, hence we cannot drop the version requirement. Unless you protest, I will close this PR

masap:
Ok, I will drop this.

Issue #219, opened 2022-02-09
Mirror port


l-rossetti:
Is it possibile to inject a mirror (span) port into a pod? Can't find any reference in code or documentation. Thanks

phoracek:
It is not possible with the current implementation. Having OVS' port mirroring available in the CNI would  be nice, do you have an idea of how the API could look like?

l-rossetti:
To create a mirror it could be done in two places:
- at NetworkAttachmentDefinition level, so that all the ports of a specific vlan are mirrored: `mirror: "mirror0"`
- at pod level, so that the specific port of this pod is mirrored: `mirror: "mirror0"`

While to inject the mirror interface we may need on the destination pod an annotation like `bind_to_mirror: "mirror0"`

Do you think it's reasonable?

phoracek:
From the two options, I think it should be on the NetworkAttachmentDefinition level. This object is managed by an administrator while Pod belongs to untrustworthy users (sorry users). If we exposed this option on Pod, users could abuse it and force their traffic to be mirrored where it should not. Having this on NetworkAttachmentDefinition sounds good.

Could be the `bind-to-mirror` attribute part of the NetworkAttachmentDefinition too? For the same reasons as described in the paragraph above.

---

One remaining question is about scheduling - I assume that mirroring is limited to the node, right? So somebody would need to make sure that both the producer and consumer pods are on the same node. We can make this easy for ourselves by saying that this is a responsibility of the owner of the consumer pod - they have to set their podAffinity to land on the same node as the producer. Would that be sufficient?

l-rossetti:
>From the two options, I think it should be on the NetworkAttachmentDefinition level. This object is managed by an administrator while Pod belongs to untrustworthy users (sorry users). If we exposed this option on Pod, users could abuse it and force their traffic to be mirrored where it should not. Having this on NetworkAttachmentDefinition sounds good.

I partially agree with this concern. Allowing users to mirror the pod traffic with an annotation on the pod itself can be an issue only in terms on waste of resources, but is not adding any security risk. The risk is more on allowing consumers to read mirrored traffic. Moreover having the annotation only at NetworkAttachmentDefinition means that the whole VLAN would be mirrored even tough you may want to mirror just the traffic of a subset of pods/ports.  
Anyway I believe that starting with NetworkAttachmentDefinition can be already a good starting point.

[...]

PR #109, opened 2020-03-03
Purge ports without interfaces

fhofherr:
As described in GH-106 a node failure leads to the OVS CNI driver's DEL
operation being called without the CNI_NETNS argument. This made the
CNI driver panic.

An empty CNI_NETNS argument is allowed according to version 0.4.0 of the
CNI spec
(https://github.com/containernetworking/cni/blob/spec-v0.4.0/SPEC.md).

The spec suggests to "clean up as many resources as possible
(e.g. releasing IPAM allocations) and return a successful response". In
accordance with this suggestion the OVS CNI driver now looks for
interfaces with an error value set. It then iterates over the ports
belonging to those interfaces and removes them.

PR #227, opened 2022-03-11
OVS Traffic Mirroring Proposal

l-rossetti:
This is a proposal for supporting the traffic mirroring feature of OVS into ovs-cni plugin. The topic has been initially discussed in [issue 219](https://github.com/k8snetworkplumbingwg/ovs-cni/issues/219).
The goal is to create and manage multiple mirror ports through either ovs-cni or a dedicated cni-plugin in Network Configuration List (Multus) as defined in CNI spec 0.4.0.

```release-note
Added support for port mirroring
```


docs/traffic-mirroring-proposal.md
```diff
@@ -0,0 +1,1282 @@
...
+Instead, in cni-args approach, you can pass mirror configuration via Deployment and use NADs only to define OVS ports and VLANs. Anyway, in most of the cases, pods in the same VLAN ID will produce to one set of mirrors. From a developer perspective, cni-args is much more flexible with respect to NAD approach. 
+
+From an isolation point of view, NAD seems more secure: it limits the resources (the mirrors in this case) a pod can create/use. While with cni-args any pod may potentially request to produce/consume to/from any mirror without constraints. Although, we are not expert in kubernetes resources security/isolation, so the disadvantage we highlighted above may not be a real issue.
+
+It's also possible to create 2 different plugins:
+
+- `ovs-mirror-producer`: it only adds input ports (previously created by `ovs` plugin) into `select_*` columns of `Mirror` table
+- `ovs-mirror-consumer`: it only adds the output into `output_port` and `output_vlan` columns of `Mirror` table
+
+The advantage of having 2 different plugins is an easier input validation, in fact we can define specific JSON keys for a better developer experience.
```
phoracek, line 1073: I would be all for introducing `ovs-mirror` or `ovs-mirror-producer` and `ovs-mirror-consumer` as independent executables that are kept under a single project, reusing existing libraries. It would may be a good start to make this project more extendable.

docs/traffic-mirroring-proposal.md
```diff
@@ -0,0 +1,1282 @@
...
+                            "name": "mirror-2",
+                            "output": "port"
+                        }
+                    ]
+                }
+            }
+          ]
+```
+
+### Pros and Cons - NAD approach VS cni-args approach
```
phoracek, line 1060: Security is the highest priority, user experience is close second though.

Would it make sense to start by configuring this through the NAD. And as a follow-up introduce a parameter `allowPodArgs` (or something similar) that would enable users to configure the mirror through parameters on pods? That way the admin has full control and option to loosen the security.

phoracek (commented):
Thanks @l-rossetti @Ks89, I love that this has plenty of examples and pictures. I'm leaving an initial review. I'm hopping to pull somebody else from my team to look at this and provide a second opinion

docs/traffic-mirroring-proposal.md
```diff
@@ -0,0 +1,1282 @@
...
+}
+```
+
+`BRIDGE_NAME`: string that represents the unique name of the bridge in ovs database where the mirror should be added
+
+`MIRROR_NAME`: string that represents the unique name of the mirror in ovs database
+
+`SELECTION_ARRAY`: array of strings that represents the type of selections. Allowed values are:
+- `src_port`
+- `dst_port`
```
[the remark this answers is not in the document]
l-rossetti: totally agree with you :D

docs/traffic-mirroring-proposal.md
```diff
@@ -0,0 +1,1282 @@
...
+                "bridge": "br1",
+                "vlan": 200
+            },
+            {
+                "type": "ovs-mirror",
+                "bridge": "br1",
+                "mirrors": [
+                    {
+                        "name": "mirror-2",
+                        "selection": ["src_port"]
```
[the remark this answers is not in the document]
l-rossetti: right, it's a typo

docs/traffic-mirroring-proposal.md
l-rossetti: Totally agree with you about security!
Your idea seems very reasonable, just need to understand how to use this `allowPodArgs`

docs/traffic-mirroring-proposal.md
```diff
@@ -0,0 +1,1282 @@
...
+            }
+        ],
+        "dns":{}
+    }
+}
+```
+
+## Questions
+
+- Who is responsible for the mirror creation?
```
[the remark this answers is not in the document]
l-rossetti: Totally agree, for now we prefer the third option.

docs/traffic-mirroring-proposal.md
```diff
@@ -0,0 +1,1282 @@
...
+    - both producer and consumer could operate in INSERT OR UPDATE mode
+- Who is responsible for the mirror deletion?
```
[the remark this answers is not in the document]
l-rossetti: ok, seems reasonable. Could you please point out the code line cleaning up the orphaned ports?

docs/traffic-mirroring-proposal.md
```diff
@@ -0,0 +1,1282 @@
...
+- How does OVSDB handles concurrency? E.g. a producer and a consumer (or 2 producers) try to insert the same row in the `Mirrors` table at the same time
```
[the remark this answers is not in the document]
l-rossetti: Agree, we have your same doubt. Let's see if someone else has a better understanding about that.

docs/traffic-mirroring-proposal.md
```diff
@@ -0,0 +1,1282 @@
...
+  name: ovs-vlan200-cons-mir1
+spec:
+  config: |-
+    {
+        "cniVersion": "0.4.0",
+        "plugins": [
+            {
+                "type": "ovs",
+                "bridge": "br1",
+                "vlan": 200
```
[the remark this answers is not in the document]
l-rossetti: You are right. In practice there's no point in setting the VLAN ID for the mirror output port.

docs/traffic-mirroring-proposal.md
```diff
@@ -0,0 +1,1282 @@
...
+
+`BRIDGE_NAME`: string that represents the unique name of the bridge in ovs database where the mirror should be added
+
+`MIRROR_NAME`: string that represents the unique name of the mirror in ovs database
+
+`SELECTION_ARRAY`: array of strings that represents the type of selections. Allowed values are:
+- `src_port`
+- `dst_port`
+- `all`
+- `vlan`
```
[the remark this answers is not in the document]
l-rossetti: Probably we should only consider dst_port and src_port, since with these two options you can build the most fine grained configuration for the mirrors. The only important aspect is to allow the users to set both src and dst at the same time.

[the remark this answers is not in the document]
phoracek: I see, thanks

docs/traffic-mirroring-proposal.md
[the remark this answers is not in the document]
phoracek: It was introduced via https://github.com/k8snetworkplumbingwg/ovs-cni/pull/109/files. We could have cleanMirrors too.

phoracek:
Hey @l-rossetti, just wanted to check - is there any new development on this? Any way I can help?

l-rossetti:
> Hey @l-rossetti, just wanted to check - is there any new development on this? Any way I can help?

Hi @phoracek! We just updated the PR based on the discussion. Could please review the proposal again? If it looks ok, we can draft a first implementation of the two plugins: `ovs-mirror-producer`, `ovs-mirror-consumer`.
If you agree, we can start by focusing only on the `cmdAdd`, leaving `cmdCheck` and `cmdDel` for a second round. 

We still have a doubt about the `ingress` and `egress` keywords: in the NAD should we consider ingress and egress from the point of view of OVS or the point of view of the Pod? 
OVS-DB specification states that:
```
select_dst_port: set of weak reference to Ports
     Ports on which departing packets are selected for mirroring.
select_src_port: set of weak reference to Ports
     Ports on which arriving packets are selected for mirroring.
 ```
But here we could take the point of view of the pod if it makes sense.

Moreover, where do you suggest to put the code? Is it ok if we create two public repos in our organization with the initial draft? Do you suggest to start with only one repo for both consumer and producer, or two dedicated repo? 
Anyway, please, feel free to suggest what you think is more suitable to everyone.

**Comment by phoracek** [comment]
Apologies for my late reply.

Thanks for the update. I left only some suggestions to fix formatting, but the proposal sounds good to me.

If you find ingress and egress problematic, we can stick to OVS' naming, that works fine.

About the code, for me the ideal process would be following:

1. You can create a quick PoC, ignoring the code quality (use bash if it helps you get it done faster), just to get a demo running to see if it will all work and get a feeling for it.
2. Extend this repository with the new functionality. Unlike the PoC, this will require well organized code, e2e tests, documentation. We can split this step to cmdAdd, cmdDel, cmdCheck. Or even start by implementing just a subset of the API. The only condition is that after each merge, we will be able to exercise and test the code.

I think that it would be useful to keep the two new CNIs under this project, to leverage the existing infrastructure. But if you'd prefer to keep it outside, I think that at least keeping both consumer and producer under the same roof would be better over splitting it.

That being said, if you'd prefer a different workflow, I'm totally happy to discuss that.

**Comment by phoracek** [comment]
This looks great! 

About the cleanup issue, would it help if we called the cleanup function on the beginning of every cmdAdd too? To make sure we start with a clean state. We should do that in both OVS CNI and the mirror CNI. If a node was forcefully restarted, DEL would not be called, so we should do it in ADD.

As for why it is not executed even in DEL, we execute the cleanup function only when something does not go as expected. If DEL progresses well, we should remove all the ports explicitly. Only if there is an issue during the DEL command, we resort to best effort cleanup using cleanPorts https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/pkg/plugin/plugin.go#L519. However, that means that leftovers from other contains would not get cleaned up. Perhaps we could unconditionally call cleanPorts on every DEL.

I don't know why CRI is not running DEL, but I would suspect that they rely on ADD being transactional - if we fail ADD, we should clean up after ourselves.

I can't think of a good solution for the convertToArray issue.

I did a high level review and it looks pretty straightforward. I have just a  few notes/questions:

* Why does the consumer need previous interface? It makes sense for producers, where we want to sniff the traffic of the previous interface. But consumer interface should be not dependent on the previous one, no? I'd expect that I can create a consumer without connecting the container to the bridge https://github.com/auroralabs-eu/ovs-cni-mirror/blob/master/pkg/consumer/consumer.go#L94
* Is it not possible to have more than one consumer on a mirror? https://github.com/auroralabs-eu/ovs-cni-mirror/blob/master/pkg/consumer/consumer.go#L106

**Comment by l-rossetti** [comment]
> About the cleanup issue, would it help if we called the cleanup function on the beginning of every cmdAdd too? To make sure we start with a clean state. We should do that in both OVS CNI and the mirror CNI. If a node was forcefully restarted, DEL would not be called, so we should do it in ADD.

Yes, we agree it is better to always call it at every ADD, knowing the state is always the cleanest possible. The thing is that the cleanup should be invoked only by ovs-cni because it involves ports/interfaces, while the ovs-mirror plugin should treat only mirrors and should attach the ports created by ovs-cni to the mirror itself. Is shouldn't be its responsibility to delete ports/interfaces that it didn't create.

> As for why it is not executed even in DEL, we execute the cleanup function only when something does not go as expected. If DEL progresses well, we should remove all the ports explicitly. Only if there is an issue during the DEL command, we resort to best effort cleanup using cleanPorts https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/pkg/plugin/plugin.go#L519. However, that means that leftovers from other contains would not get cleaned up. Perhaps we could unconditionally call cleanPorts on every DEL.

Ok , we understand the logic you put in place. But yes, as before, it would be much better to call the cleanup every DEL of ovs-cni. This would simplify also the logic built around the cleanup caller.

> I don't know why CRI is not running DEL, but I would suspect that they rely on ADD being transactional - if we fail ADD, we should clean up after ourselves.

Probably we misunderstood the spec in case of chained plugin: I don't find an explicit sentence specifying that a DEL for the upstream plugin (e.g. ovs-cni) of the chain must be called by the CRI in case the downstream one (e.g. ovs-mirror) fails. 

On the other hand we are found this in the [plugin-delegation section](https://github.com/containernetworking/cni/blob/main/SPEC.md#section-4-plugin-delegation) of the SPEC: 
"_If, on ADD, a delegated plugin fails, the "upper" plugin should execute again with DEL before returning failure._"

So this doubt comes to our minds: are we sure ovs-cni-mirror should be implemented as a dedicated plugin instead of a delegated plugin (as done for the IPAM one)? Cause actually the ovs-cni-mirror will never be called by itself but it will be always in chain with ovs-cni, cause is responsibility of the latter to create the interface/port. The sentence of the spec I quoted about would allow us to avoid the call  to the cleanup function in case of ovs-mirror failure.
What do you think?

> Why does the consumer need previous interface? It makes sense for producers, where we want to sniff the traffic of the previous interface. But consumer interface should be not dependent on the previous one, no? I'd expect that I can create a consumer without connecting the container to the bridge https://github.com/auroralabs-eu/ovs-cni-mirror/blob/master/pkg/consumer/consumer.go#L94

It is needed because the consumer port/interface created by ovs-cni must be added as "output_port" in the mirror table. So we need the previousResult to be able to fetch the port reference and add it to the output_port column of the table [here](https://github.com/auroralabs-eu/ovs-cni-mirror/blob/2268aa4634401e471cdd8bb2d319ce5e872991bd/pkg/consumer/consumer.go#L63).

> Is it not possible to have more than one consumer on a mirror? https://github.com/auroralabs-eu/ovs-cni-mirror/blob/master/pkg/consumer/consumer.go#L106

Unfortunately not, cause [OvsDB-Schema manual](http://www.openvswitch.org//ovs-vswitchd.conf.db.5.pdf) specifies at page 41 that "output_port" is an "_optional weak reference to Port_", not a "_Set of weak references to Port**s**_". 

We see only two different options to have more than one consumer:
- create one mirror for each consumer and allow the other pods to produce to all the mirrors
- implement ovs-cni-mirror to support "output_vlan" that is "_The frames will be sent out all ports that trunk output_vlan, as well as any ports with implicit VLAN output_vlan. When a mirrored frame is sent out a trunk port, the frame’s VLAN tag will be set to output_vlan, replacing any existing tag; when it is sent out an implicit VLAN port, the frame will not be tagged. This type of mirroring is sometimes called RSPAN_."

About the latter option we excluded it in our proposal cause we wanted to start fulfilling the requirements of our project and we don't have enough knowledge about RSPAN. For sure this would be a nice to have improvement for the future.

**Comment by Ks89** [comment]
We have a question about UUIDs, because [here](https://github.com/auroralabs-eu/ovs-cni-mirror/blob/2268aa4634401e471cdd8bb2d319ce5e872991bd/pkg/ovsdb/ovsdb.go#L595) we are using the mirrorName as done in ovs-cni with portName.
However, mirrorName is a value passed by the user via NAD (also with special characters like "-" which is accepted as mirror-name by ovs), so it must be validated and cleaned to be used as "Operation.UUIDName".
Instead, in ovs-cni this is not required, because portName is generated from interfaceName.

How could we create a valid UUIDName for mirrors? For example we can create a random string with a fixed length and without special characters. Do you have a better solution?
Thanks

**Comment by l-rossetti** [comment]
Hi @phoracek how are you doing? Any comment about the previous messages?

**Comment by phoracek** [comment]
Hello,

> Ok , we understand the logic behind. But yes, as before, we may check the state of the db at each ovs-cni invocation. This would simplify also the logic built around the cleanup caller in ovs-cni. On the other side may seem overkill.

If this simplifies the code, I don't mind. I don't imagine it would have a terrible performance impact since we access the OVS database using the unix socket.

> So this doubt comes to our minds: are we sure ovs-cni-mirror should be implemented as a dedicated plugin instead of a delegated plugin (as done for the IPAM one)? Cause actually the ovs-cni-mirror will never be called by itself but it will be always in chain with ovs-cni, cause is responsibility of the latter to create the interface/port. The sentence of the spec I quoted above would allow us to avoid the call to the cleanup function in case of ovs-mirror failure. But in that case there are some other impacts to evaluate.

In case of IPAM, the IPAM CNI returns the IP that should be configured. And then the actual IP assignment is done by the main CNI. That is unlike mirror CNI where the configuration is actually applied by the CNI.

The reference plugins under https://github.com/containernetworking/plugins/tree/main/plugins/meta are all relevant only when chained with some other plugin, like mirror. Those plugins probably suffer the same cleanup issue. Maybe we should raise that limitation as a bug of CNI, so it could be resolved for all?

I'm not totally against the idea of using delegates, but we should have a good reason to break this pattern.

> About the latter option we excluded it in our proposal to fulfill the requirements of our project and we don't have enough knowledge about RSPAN. For sure this would be a nice to have improvement for the future.

+1 on keeping it simple for now with the first option.

> How could we create a valid UUIDName for mirrors? For example we can create a random string with a fixed length and without special characters. Do you have a better solution?

Generating a random UUID sounds fine. I assume we can then keep the mirror name as other-config, or something similar?

If I missed any other question, please don't hesitate to bring it up.

**Comment by Ks89** [comment]
Hi @phoracek 
following [your suggestion  here](https://github.com/k8snetworkplumbingwg/ovs-cni/pull/227#issuecomment-1106604897) we merged [this repo](https://github.com/auroralabs-eu/ovs-cni-mirror) into ovs-cni.

The main changes with respect to the previous commit are:
- upgrade go to 1.18 to use generics
- integrate `MirrorNetConf` in `types.go` and use it in `config.go` with GO generics to avoid code duplication
- change "insert" operation in `Mirror` table with a fixed transaction `uuidName` [HERE](https://github.com/auroralabs-eu/ovs-cni/blob/d955a45c3976ad52056f3083bb2377c6d4994cd3/pkg/ovsdb/ovsdb.go#L871), because `named-uuid` represents a reference to inserted rows and not a real UUID as explained in [RFC7047](https://datatracker.ietf.org/doc/html/rfc7047.txt)
- as discussed previously, we removed `cleanPorts` from ovs-cni-mirror plugins and added both at the beginning of `cmdAdd` and at the end of `cmdDel` of ovs-cni plugin 

To discuss:
- integrate ovs-cni-mirror in your CI/CD, please could you provide more information about this topic?
- do you think that the documentation `traffic-mirroring.md` is complete enough?
- what is missing in this PR to be eligible for the merge?

**Comment by phoracek** [comment]
>     * integrate ovs-cni-mirror in your CI/CD, please could you provide more information about this topic?

We should get at least a silver bullet test verifying that the solution works into our e2e test suite, running on Kubernetes. All that should be required is to introduce a new module under tests/, similar to https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/tests/ovs_test.go.

Rest of the test cases may be covered in integration tests, similar to this one: https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/pkg/plugin/plugin_test.go. These can run without Kubernetes and are considerably faster to execute.

>     * do you think that the documentation `traffic-mirroring.md` is complete enough?

I like the contents of the docs, thanks for pictures and plenty of examples! The only extra thing I'd ask for is a sentence or two about each of the cases. I may have some minor formatting comments once I get to review it in detail.

>     * what is missing in this PR to be eligible for the merge?

Tests and thorough review. Once you post the test coverage, I would try it out locally and then review the whole PR. I skimmed through the code and it seems that wast majority of the changes only add new code and modules, leaving existing code intact, so I don't anticipate issues there.

**Comment by phoracek** [comment]
@Ks89 tests are green now \o/ let me know when it will be ready for a review

**Comment by Ks89** [comment]
> @Ks89 tests are green now \o/ let me know when it will be ready for a review

ok, thanks.
I need to write some other tests, but I hope to complete everything next week.

**Comment by Ks89** [comment]
Hi @phoracek,
we added unit tests and e2e tests as requested and improved a little bit the code.

Some point to notice:
- we added a new parameter to `CreatePrivilegedPodWithIP` to pass an additional command when starting up the pod. This is required to run `tcpdump`, because we tried by running it in background `tcpdump ... &` but it was returning error `exit code 1`. There is a known issue also discussed [here](https://github.com/kubernetes/kubernetes/issues/44371) and none of those solutions worked.
- we have no idea why running unit tests on your infrastructure returned an error. If we run `make test` on our local machine, everything looks fine: 76 tests passed with no errors.
- we have no idea why running e2e tests with cni version 0.3.1 we receive the error below on `clusterApi.DeletePodsInTestNamespace()`. It seems a multus error. If we try the same for both 0.3.0 and 0.4.0 everything works as expected. Also, If we remove `DeletePodsInTestNamespace` and we use `CreateTestNamespace` and `RemoveTestNamespace` every test it seems to be ok. Why? We noticed that current multus version in kubevirtci is 3.4.2, but the latest is 3.9. Is there a way to update it?

```
Warning  FailedKillPod  5s (x11 over 2m12s)  kubelet            error killing pod: failed to "KillPodSandbox" for "67e847f2-1a74-43c3-820b-c6a33e1df363" with KillPodSandboxError: "rpc error: code = Unknown desc = failed to destroy network for pod sandbox k8s_pod-prod-2_test-namespace_67e847f2-1a74-43c3-820b-c6a33e1df363_0(2bc2f7c62a00a1d5f9729d3436be2d9f754c6cb665526993ebb8a51356823402): error removing pod test-namespace_pod-prod-2 from CNI network \"multus-cni-network\": delegateDel: error invoking ConflistDel - \"ovs-net-prod\": conflistDel: error in getting result from DelNetworkList: netplugin failed with no error message"
```

**Comment by phoracek** [comment]
Awesome, thanks! I bumped the version of Multus via https://github.com/k8snetworkplumbingwg/ovs-cni/pull/238. I'll go over the PR and the rest of your points during this week.

/retest

**Comment by phoracek** [comment]
Running the tests locally, I noticed that removal of pods using the new CNI fail due to nil pointer dereference. That causes them to get stuck terminating.

```
  Warning  FailedKillPod   45s               kubelet            error killing pod: failed to "KillPodSandbox" for "53299780-60c3-4a98-94bd-481bd3b10f5e" with KillPodSandboxError: "rpc error: code = Unknown desc = failed to destroy network for pod sandbox k8s_pod-prod-2_test-namespace_53299780-60c3-4a98-94bd-481bd3b10f5e_0(7f744c750ec5aa5778cac31419d701f90c66ee660792d6b96c7d326637f8244b): error removing pod test-namespace_pod-prod-2 from CNI network \"multus-cni-network\": delegateDel: error invoking ConflistDel - \"ovs-net-prod\": conflistDel: error in getting result from DelNetworkList: netplugin failed: \"2022/08/19 12:13:19 CNI DEL was called for container ID: 7f744c750ec5aa5778cac31419d701f90c66ee660792d6b96c7d326637f8244b, network namespace /var/run/netns/0d7b05ac-6229-4e18-8d07-c1a4b2d833b3, interface name net1, configuration: {\\\"bridge\\\":\\\"br-test\\\",\\\"cniVersion\\\":\\\"0.3.0\\\",\\\"mirrors\\\":[{\\\"egress\\\":true,\\\"ingress\\\":true,\\\"name\\\":\\\"mirror-1\\\"}],\\\"name\\\":\\\"ovs-net-prod\\\",\\\"type\\\":\\\"ovs-cni-mirror-producer\\\"}\\npanic: runtime error: invalid memory address or nil pointer dereference\\n[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x64341a]\\n\\ngoroutine 1 [running, locked to thread]:\\npanic({0x66e240, 0x893b40})\\n\\t/home/phoracek/code/ovs-cni/build/_output/bin/go/src/runtime/panic.go:941....
```

**Comment by Ks89** [comment]
Ok, tests are ok now.

File: pkg/ovsdb/ovsdb.go [hunk 10]
```diff
@@ -329,7 +336,302 @@ func (ovsd *OvsDriver) GetOFPortVlanState(portName string) (string, *uint, []uin
...
+		// select_src_port = Ports on which arriving packets are selected for mirroring
+		conditionIngress := ovsdb.NewCondition("select_src_port", ovsdb.ConditionIncludes, portUUID)
+		conditions = append(conditions, conditionIngress)
+	}
+	if egress {
+		// select_dst_port = Ports on which departing packets are selected for mirroring
+		conditionsEgress := ovsdb.NewCondition("select_dst_port", ovsdb.ConditionIncludes, portUUID)
+		conditions = append(conditions, conditionsEgress)
+	}
+	// We cannot call findByCondition because we need to pass an array of conditions
```
phoracek, line 529: Could you define a helper function to do that? So we don't need to duplicate this code here and in the following two functions

pkg/mirror-producer/producer_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"testing"
+)
+
+func TestPlugin(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Plugin Suite")
```
phoracek, line 26:
```suggestion
	RunSpecs(t, "Mirror Producer Suite")
```

pkg/mirror-consumer/consumer_suite_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"testing"
+)
+
+func TestPlugin(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Plugin Suite")
```
phoracek, line 26:
```suggestion
	RunSpecs(t, "Mirror Consumer Suite")
```

pkg/mirror-producer/producer_test.go
```diff
@@ -0,0 +1,888 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	plugin "github.com/k8snetworkplumbingwg/ovs-cni/pkg/plugin"
+
+	"github.com/k8snetworkplumbingwg/ovs-cni/pkg/types"
+)
+
+type MirrorNet040 struct {
```
phoracek, line 41: Please add a comment explaining why we treat 0.4.0 separatelly

pkg/mirror-producer/producer_test.go
```diff
@@ -0,0 +1,888 @@
...
+	output, err := exec.Command("ovs-vsctl", "get", "Port", name, "_uuid").CombinedOutput()
+	if err != nil {
+		return "", fmt.Errorf("failed to get port uuid by name: %v", string(output[:]))
+	}
+
+	return strings.TrimSpace(string(output[:])), nil
+}
+
+// function to get a mirror attribute
+func getMirrorAttribute(mirrorName, attributeName string) (string, error) {
```
phoracek, line 786: I see that many of these helper functions are duplicated in consumer_test. Please create a test helper package for mirror plugins

pkg/mirror-producer/producer_test.go
```diff
@@ -0,0 +1,888 @@
...
+
+				By("run ovs-cni-mirror-producer passing prevResult")
+				confMirror1, result1 := testAdd(conf, mirrors, prevResult1, IFNAME1, targetNs)
+				confMirror2, result2 := testAdd(conf, mirrors, prevResult2, IFNAME2, targetNs)
+				testCheck(confMirror1, result1, IFNAME1, targetNs)
+				testCheck(confMirror2, result2, IFNAME2, targetNs)
+
+				// mirror must have both ports (hostIfaces obtained from 'prevResult1' and 'prevResult2')
+				By("verify that mirrors have all ports")
+				checkPortsInMirrors(mirrors, prevResult1, prevResult2)
```
phoracek, line 519: Is not this already asserted by the two testAdd calls above?

pkg/mirror-producer/producer_test.go
```diff
@@ -0,0 +1,888 @@
...
```
phoracek, line 580: ditto

pkg/mirror-producer/producer_test.go
```diff
@@ -0,0 +1,888 @@
...
+				prevResult := createInterfaces(IFNAME1, targetNs)
+
+				By("run ovs-cni-mirror-producer passing prevResult")
+				confMirror, result := testAdd(conf, mirrors, prevResult, IFNAME1, targetNs)
+				testCheck(confMirror, result, IFNAME1, targetNs)
+
+				By("create a consumer interface and add its port via 'ovs-vsctl' to fill mirror 'output_port'")
+				r2 := createInterfaces(IFNAME2, targetNs)
+				portUUID := getPortUUIDFromResult(r2)
+				addOutputPortToMirror(portUUID, mirrors[0].Name)
```
phoracek, line 630: Do we need this helper here? Could this be tested by adding several mirror users, removing one, confirming that the mirror still exists, and continuing until there are no users and only then confirm the mirror is gone?

pkg/mirror-producer/producer.go
```diff
@@ -0,0 +1,224 @@
...
+	netconf, err := config.LoadMirrorConf(args.StdinData)
+	if err != nil {
+		return err
+	}
+
+	ovsDriver, err := ovsdb.NewOvsBridgeDriver(netconf.BrName, netconf.SocketFile)
+	if err != nil {
+		return err
+	}
+
```
phoracek, line 94: What happens with orphaned mirrors? Say we called ADD multiple times and then rebooted the machine. Orphaned ports would be cleaned up by the main CNI. Will mirrors be cleaned up too or do we need to clean up from here?

phoracek (changes requested):
Overall this looks good, just some smaller comments. Please push new changes as commits at the top of the branch, to make reviews easier, thanks.

pkg/mirror-producer/producer_test.go
Ks89: We added this to create a more realistic test where a producer have also output_port (basically 99,99% of the times in real scenarios). cmdDel logic of mirror-producer must be able to cleanup a mirror only if also output-port is empty.
The same for mirror-consumer, because it can remove a mirror only if also select_* are empty.

Are you talking about *consumers* with "mirror users"?
If yes, the answer is no, because a mirror can have multiple producers but only a single consumer, in fact ovsdb spec states that `output_port` is an optional weak reference to Port, and not "set of weak reference...".
Instead, if you are talking about *producers*, that scenario is covered by "adding multiple ports to a single mirror", because there is a single mirror with 2 portUUIDS in select_*. However, we could improve this test checking mirror ports between the 2 cmdDel and at the end.

pkg/mirror-producer/producer_test.go
Ks89: You are right, it'a good idea.
Also plugin_test.go doesn't have a comment about this (Net040 and NetCurrent).
We should write a comment and add it to both plugin, mirror_producer and mirror_consumer files.
Do you have a suggestion? The difference is about `PrevResult` type, but how can we explain this in a clear way?

Ks89:
We fixed some stuff, but it's not completed. We'll continue next Monday.
We added some comments to ask more information in a couple of requests.

cmd/mirror/consumer/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal
```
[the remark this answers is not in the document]
Ks89: Yes, no problem.

pkg/mirror-producer/producer_test.go
Ks89: I created `/pkg/testhelpers/mirror.go` with package `testhelpers`.
Do you prefer this location or inside `/tests/`?

pkg/ovsdb/ovsdb.go
```diff
@@ -329,7 +336,302 @@ func (ovsd *OvsDriver) GetOFPortVlanState(portName string) (string, *uint, []uin
...
+		operations := []ovsdb.Operation{*mirrorOp, *attachMirrorOp}
+
+		_, err = ovsd.ovsdbTransact(operations)
+		return err
+	}
+	return nil
+}
+
+// DeleteMirror Removes a mirror of a specific bridge
+func (ovsd *OvsBridgeDriver) DeleteMirror(bridgeName, mirrorName string) error {
```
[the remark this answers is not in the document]
Ks89: agree with you. Done

pkg/ovsdb/ovsdb.go
Ks89: agree. done

tests/cluster/cluster.go
```diff
@@ -187,6 +187,16 @@ func (api *ClusterAPI) PingFromPod(podName, containerName, targetIP string) erro
 	return nil
 }
 
+// ReadTCPDumpFromPod run the cat command on the pod container to read the content of /tcpdump.log
+func (api *ClusterAPI) ReadTCPDumpFromPod(podName, containerName string) (string, error) {
```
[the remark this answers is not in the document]
Ks89: yes, you are right

pkg/mirror-producer/producer_test.go
phoracek: It can be as simple as the explanation here I think: https://github.com/k8snetworkplumbingwg/ovs-cni/pull/227/files#diff-14c657ea5956b13b208b921905d0d631a1835176df2a42bd4a3f1bb42bdae160R160

add prevResult from cache, because missing in CNI spec < 0.4.0

pkg/mirror-producer/producer_test.go
phoracek: It's fine under pkg, at least it does not mix integration and e2e tests

pkg/mirror-producer/producer_test.go
phoracek: Actually now when I look at it again, I don't think we need that at all... I was confused before thinking that this is related to prevresult being related from cache.

pkg/mirror-producer/producer_test.go
phoracek: Ah, thanks, I think I understand now

pkg/mirror-producer/producer_test.go
Ks89: Ok, should I remove the comment here
```
// MirrorNet040 struct that represent the network configuration for CNI spec 0.4.0
type MirrorNet040 struct {
	...
}

// MirrorNetCurrent struct that represent the network configuration for CNI spec 1.0.0
type MirrorNetCurrent struct {
	...
}
```

or can I leave it?

Ks89:
Now there are these two errors:
```
/tmp/src/github.com/k8snetworkplumbingwg/ovs-cni/.gopath/src/github.com/k8snetworkplumbingwg/ovs-cni/pkg/testhelpers/mirror.go:13:2: should not use dot imports
/tmp/src/github.com/k8snetworkplumbingwg/ovs-cni/.gopath/src/github.com/k8snetworkplumbingwg/ovs-cni/pkg/testhelpers/mirror.go:14:2: should not use dot imports
```
I copied this code 
```
	. "github.com/onsi/ginkgo"
	. "github.com/onsi/gomega"
```
from plugin_test.go.

Why is it a problem in testhelpers/mirror.go and not in plugin_test.go? Probably it's related to the file name "_test.go". How can we fix this? Should we use `testhelpers/mirror_test.go` or skip that file in the golint command?

pkg/mirror-producer/producer.go
phoracek: I mean ports in OVS DB that were created for veth interface connecting containers. After restart, these ephemeral interfaces are gone, byt the OVS DB entries are still there. These entries will be deleted on the first call of `ovs` ADD. When all these are removed, it could mean that a previously existing mirror does not have any producers anymore and should be cleaned up.

Let me try to come up with a sequence:
1. container1 was attached to br1 through port1, producing to mirror1
2. container2 was attached to br2 through port2, producing to mirror2
3. host is powered off and on
4. host comes up, OVS DB has all the previously created ports and mirrors, but interfaces are not found in the system
5. container1 is scheduled back on this node, container2 is not
6. container1 calls `ovs` CNI ADD, this call runs a cleanup of orphaned port1 and port2, creating a new port3, producing to mirror1

Now mirror 2 is unused but still in the DB. Should `ovs-mirror-producer` CNI call cleanup on ADD, so container1 attachment would clean up mirror2?

pkg/mirror-producer/producer_test.go
phoracek: Let's leave it as is :)

Ks89:
yes no problem, I can remove the dots

phoracek:
Oh :D sorry, I edited your message instead of responding to it...

pkg/mirror-producer/producer.go
Ks89: Ok, now it's clear.
Ideally speaking yes, with a host shutdown this could be the right behaviour, however what will happen in case of a restart of a pod? This will call ADD and it could trigger that behaviour also in a normal scenario.

In ovsdb it's possibile to create a mirror without both producer and consumer and this is exactly what we do in plugins, because both plugins (producer and consumer) try to add a new mirror if not already existing. Next, they try to configure that mirror adding select_* and output_port.
This worked very well in case of multiple pods started at the same time, because concurrent operation are handled very easely.
This means that in a specific moment a mirror could exist without ports. Adding a logic to remove mirrors without ports could cause an unexpected behaviour.

How could ovs-mirror-producer identify mirrors to remove, if ports are cleaned-up by ovs-cni?? Because, It can find all mirrors without ports and remove them. However, how do it know if those are mirrors with errors or not? It's not like the relationship of interfaces and ports. In that case, you can search for interfaces with errors/unexisting, then search related ports and clean up everything.

Also, mirrors could be used to have SPAN ports for SIEM or other security software. It could be dangerous to remove a mirror configured in a plugin from another one. Let me explain:
1) ovs-mirror-* plugin creates mirror 1
2) ovs-mirror-* plugin creates mirror 2
3) shutdown
4) ovs-mirror-* plugin (the one used to create mirror 1)  see the mirror 2 is useless and it removes mirror 2
The last step is quite confusing. Why a plugin configured to operate on mirror-1 is removing mirror-2?
With interfaces its quite easy to find the error condition, but with mirrors it's not so easy. 
We think that ovs-mirror plugin should interact with `Mirror table` only on the specific row (identified by the mirror name passed via NAD). Mirrors are used to read network traffic and a mistake could cause security issues.


But there is another case where the `Mirror` table could become inconsistent:
if you remove a port (either via `cleanPorts` or manually), its UUID reference won't be removed from `Mirror` table (`output` and `select_*` columns).
There are different solutions:
1) ignore the problem
2) when you remove a port via `cleanPorts`, you also need to search and remove it in all mirrors in `Mirror` table, but it means that `ovs-cni` will interact with `Mirror` table even if it shouldn't be its duty
3) when both `ovs-mirror-producer` and `ovs-mirror-consumer` add or delete a port UUID they should also check the existance of all UUIDs referenced in the `Mirror` table for that specific mirror (and not other mirrors, for the same security reason described above)

Ks89:
> Oh :D sorry, I edited your message instead of responding to it...

updated removing the dot imports and fixed a stupid bug in e2e tests after renaming ovs-cni-mirrror-* into ovs-mirror-*.

Also added a big comment to the discussion above https://github.com/k8snetworkplumbingwg/ovs-cni/pull/227#discussion_r960739841

pkg/mirror-producer/producer.go
Ks89: Hi @phoracek do you have any comments about this?

phoracek: Hey, sorry about the delay, I was on a vacation last week. I will re-review this this week

phoracek: > This means that in a specific moment a mirror could exist without ports. Adding a logic to remove mirrors without ports could cause an unexpected behaviour.

Would it be possible create the mirror and attach the first port it in a single transaction? That may fix this issue with transiently unused mirror.

> How could ovs-mirror-producer identify mirrors to remove, if ports are cleaned-up by ovs-cni?? Because, It can find all mirrors without ports and remove them.

I would limit the condition of ovs-mirror-producer/consumer to "are there any ports attached to the mirror". If there are none, we can delete the mirror?

> Why a plugin configured to operate on mirror-1 is removing mirror-2?

It is confusing, but since we don't have any central daemon that could oversee all objects we created, we don't have another choice. It would be consistent with the cleanup done in ovs plugin itself - when invoked to add a port, it would also remove all orphaned ports connected to other bridges it has nothing to do with.

> We think that ovs-mirror plugin should interact with Mirror table only on the specific row (identified by the mirror name passed via NAD).

I see. But this would not allow us to clean up after Pods that after restart got scheduled to a different node, no? Is it possible to mark mirrors managed by the CNI, so we don't touch what's not ours (similar to port cleanup). We created those mirrors, so it should be ok if we remove them - or would it cause issues integrating with other software.

> when both ovs-mirror-producer and ovs-mirror-consumer add or delete a port UUID they should also check the existance of all UUIDs referenced in the Mirror table for that specific mirror (and not other mirrors, for the same security reason described above)

I would be in favor of this one.

---

In general I'm ok with the PR as is. I won't press for the cleanup to be resolved here, since it does not affect the rest of OVS CNI. I am happy to discuss it further though, your call o/

Ks89: Hi @phoracek 

> Would it be possible create the mirror and attach the first port it in a single transaction? That may fix this issue with transiently unused mirror.

In practice it seems feasible, however we need to do some experiments with transaction.


> It is confusing, but since we don't have any central daemon that could oversee all objects we created, we don't have another choice. It would be consistent with the cleanup done in ovs plugin itself - when invoked to add a port, it would also remove all orphaned ports connected to other bridges it has nothing to do with.

It seems that cleanPorts can remove only ports of the current Bridge, even if it finds all interfaces of all bridges with an error (https://github.com/k8snetworkplumbingwg/ovs-cni/blob/e604d234b7895b1b88c89b466e68c56a7a187b54/pkg/ovsdb/ovsdb.go#L190) 
In our opinion this current behaviour seems ok.
If it's possible to create mirrors with a single transaction, we can also add a `cleanMirrors` function with the same logic of `cleanPorts` (limited to the current bridge).


> 5. container1 is scheduled back on this node, container2 is not
> 6. container1 calls ovs CNI ADD, this call runs a cleanup of orphaned port1 and port2, creating a new port3, producing to mirror1

Following the points above, the scenario described in steps 5 and 6 won't cleanup db for both mirrors and ports until some pods are scheduled on the same node with the same bridge and operating on a mirror (for the mirror case)


> I see. But this would not allow us to clean up after Pods that after restart got scheduled to a different node, no? Is it possible to mark mirrors managed by the CNI, so we don't touch what's not ours (similar to port cleanup). We created those mirrors, so it should be ok if we remove them - or would it cause issues integrating with other software.

At the moment, it doesn't seem a big issue and we can work on it in future improvements if required.


> I would be in favor of this one.

We made some tests and this is not required, because if a port is deleted either manually or via cleanPorts, all references to this port will be automatically cleaned across the whole db.

Ks89: Hi @phoracek

> In practice it seems feasible, however we need to do some experiments with transaction.

I'm so stupid. I totally forgot that it's already implemented with transactions :) (https://github.com/auroralabs-eu/ovs-cni/blob/99b6a1b12450175a96d23a5eaeab93504c6a0134/pkg/ovsdb/ovsdb.go#L351)
So, yes, no empty mirrors will be added by the plugin.


> At the moment, it doesn't seem a big issue and we can work on it in future improvements if required.

It's not a big issue, but to be more consistent with ovs-cni we decided to add owner to Mirror table as external_ids.


We pushed some commits with these changes:
- add owner in `external_ids` of Mirror as for Ports. We used the same logic. If a `cmdDel` want to delete a Mirror not owned by ovs-cni, it will return an error.
- implement `cleanMirrors` executed in both `cmdAdd` and `cmdDel` as for `cleanPorts` in `plugin`. Furthermore, as `cleanPorts`, `cleanMirrors` can remove mirrors only if they are in the current bridge and are owned by ovs-cni.
- add some unit tests to cover cases managed by cleanMirrors function:
1. there are some empty mirrors in db owned by ovs-cni and you want to add another mirror to the same bridge -> cleanup empty mirrors
2. there are some empty mirrors in db owned by ovs-cni and you want to remove a mirror of the same bridge -> cleanup empty mirrors
3. there are some empty mirrors in db NOT owned by ovs-cni and you want to add another mirror to the same bridge -> do nothing
4. there are some empty mirrors in db NOT owned by ovs-cni and you want to remove a mirror of the same bridge -> do nothing

We added both add and delete cases to cover cleanMirrors invocation from both `cmdAdd` and `cmdDel`

pkg/testhelpers/mirror.go
```diff
@@ -220,6 +227,33 @@ func AddOutputPortToMirror(portUUID, mirrorName string) (string, error) {
...
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		if ovsPortOwner != "" {
+			// manually add owner as external_ids
+			_, err = addOwnerToMirror(ovsPortOwner, mirrorName)
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		}
+		// check existence of the empty mirrors
+		emptyMirExists, err := IsMirrorExists(mirrorName)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		gomega.Expect(emptyMirExists).To(gomega.Equal(true))
```
phoracek, line 244: If createEmptyMirror executed without an error, I think it's safe to assume it is there. This code can be removed.

pkg/mirror-producer/producer.go
phoracek: Hey, this is awesome, thanks for adding the cleanup.

phoracek (approved):

phoracek:
Thanks so much @l-rossetti @Ks89 this is really awesome. I appreciate the extra work you put into this. Kudos!

/lgtm
/approve

pkg/testhelpers/mirror.go
Ks89: In theory yes. I added this to be really sure about the state of mirror table.
If you want I can push a commit to remove it.

phoracek: If you would find spare moment to push a PR with the cleanup, that'd be nice, your call

PR #229, opened 2022-05-17
oci: Support podman 

oshoval:
Support (rootfull) podman OCI
Auto detect running OCI

Bump kubevirtci to have podman support

Note:
`automation/check-patch.packages` still has docker as default in case used with yam / dnf

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,11 +15,13 @@ export PATH := $(GOBIN):$(PATH):$(BIN_DIR)
 GOPATH = $(CURDIR)/.gopath
 ORG_PATH = github.com/k8snetworkplumbingwg
 PACKAGE = ovs-cni
+OCI_BIN ?= docker
 REPO_PATH = $(ORG_PATH)/$(PACKAGE)
 BASE = $(GOPATH)/src/$(REPO_PATH)
 PKGS = $(or $(PKG),$(shell cd $(BASE) && env GOPATH=$(GOPATH) $(GO) list ./... | grep -v "$(PACKAGE)/vendor/" | grep -v "$(PACKAGE)/tests/cluster" | grep -v "$(PACKAGE)/tests/node"))
 V = 0
 Q = $(if $(filter 1,$V),,@)
+TLS_SETTING := $(if $(filter $(OCI_BIN),podman),--tls-verify=false,)
 
 all: lint build
 
@@ -78,14 +80,14 @@ functest: $(GO)
 docker-build: $(patsubst %, docker-build-%, $(COMPONENTS))
 
 docker-build-%: build-%
-	docker build -t ${REGISTRY}/ovs-cni-$*:${IMAGE_TAG} ./cmd/$(subst -,/,$*)
+	$(OCI_BIN) build -t ${REGISTRY}/ovs-cni-$*:${IMAGE_TAG} ./cmd/$(subst -,/,$*)
 
 docker-push: $(patsubst %, docker-push-%, $(COMPONENTS))
 
 docker-push-%:
-	docker push ${REGISTRY}/ovs-cni-$*:${IMAGE_TAG}
-	docker tag ${REGISTRY}/ovs-cni-$*:${IMAGE_TAG} ${REGISTRY}/ovs-cni-$*:${IMAGE_GIT_TAG}
-	docker push ${REGISTRY}/ovs-cni-$*:${IMAGE_GIT_TAG}
+	$(OCI_BIN) push ${TLS_SETTING} ${REGISTRY}/ovs-cni-$*:${IMAGE_TAG}
+	$(OCI_BIN) tag ${REGISTRY}/ovs-cni-$*:${IMAGE_TAG} ${REGISTRY}/ovs-cni-$*:${IMAGE_GIT_TAG}
+	$(OCI_BIN) push ${TLS_SETTING} ${REGISTRY}/ovs-cni-$*:${IMAGE_GIT_TAG}
 
 dep: $(GO)
 	$(GO) mod tidy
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -15,6 +15,7 @@ export PATH := $(GOBIN):$(PATH):$(BIN_DIR)
 GOPATH = $(CURDIR)/.gopath
 ORG_PATH = github.com/k8snetworkplumbingwg
 PACKAGE = ovs-cni
+OCI_BIN ?= $(shell if hash podman 2>/dev/null; then echo podman; elif hash docker 2>/dev/null; then echo docker; fi)
```
oshoval, line 18: This assumes that in case podman / docker exists it is also the running service
later on we can improve it if needed

Makefile
```diff
@@ -15,11 +15,13 @@ export PATH := $(GOBIN):$(PATH):$(BIN_DIR)
...
 REPO_PATH = $(ORG_PATH)/$(PACKAGE)
 BASE = $(GOPATH)/src/$(REPO_PATH)
 PKGS = $(or $(PKG),$(shell cd $(BASE) && env GOPATH=$(GOPATH) $(GO) list ./... | grep -v "$(PACKAGE)/vendor/" | grep -v "$(PACKAGE)/tests/cluster" | grep -v "$(PACKAGE)/tests/node"))
 V = 0
 Q = $(if $(filter 1,$V),,@)
+TLS_SETTING := $(if $(filter $(OCI_BIN),podman),"--tls-verify=false",)
```
oshoval, line 24: We can add a condition that it is only for local registry if desired

automation/check-patch.e2e.sh
```diff
@@ -2,7 +2,7 @@
 
 # This script should be able to execute functional tests against Kubernetes
 # cluster on any environment with basic dependencies listed in
-# check-patch.packages installed and docker running.
+# check-patch.packages installed and docker / podman running.
```
qinqon, line 5: nit: Since podman has preference can we change this to:
```suggestion
# check-patch.packages installed and podman / docker running.
```

hack/common.sh
```diff
@@ -61,3 +61,14 @@ function kubevirt_version() {
...
+determine_cri_bin() {
+    if curl --unix-socket "${XDG_RUNTIME_DIR}/podman/podman.sock" http://d/v3.0.0/libpod/info >/dev/null 2>&1; then
+        echo podman
+    elif docker ps >/dev/null 2>&1; then
+        echo docker
+    else
+        echo "error: no docker or podman found"
+        exit 1
+    fi
+}
```
qinqon, line 74: Aren't we doing this at kubevirtci ? 

https://github.com/kubevirt/kubevirtci/blob/23ddfc29c03bb1ccaeb79cd204163f8108db1c36/cluster-up/cluster/ephemeral-provider-common.sh#L17-L36

qinqon (changes requested):

hack/common.sh
oshoval: yes but there it affect only cluster-up scope, not cluster-sync and the calls of the e2e / unit tests and so on

automation/check-patch.e2e.sh
oshoval: ye btw, we can even drop this change if needed, just have the word docker as a general term
wdyt ?

qinqon: Docker is not a general term, we keep like the comment or replace by something like "OCI bin" (but no one understand that)

oshoval:
Addressed comments

oshoval:
Thanks Quique
@phoracek can you please take a look ?

oshoval:
cosmetics 
ptal

oshoval:
Thanks Quique

phoracek (commented):
/approve

oshoval:
@phoracek 
Can we create please a release, so CNAO will auto bump and take it ?
Reason we need it is because when CNAO clones ovs-cni and run ovs-cni tests within CNAO
it change dir to the cloned ovs-cni, and from there runs `./cluster/ssh.sh node01 -- sudo <cmd>`
this command still has the old ovs-cni with the old kubevirtci that doesn't have CRI auto detect
so running podman on CI will fail

```
./cluster/ssh.sh node01 -- sudo ovs-vsctl add-br br-test
stdout: ...
, stderr Cloning into '/tmp/deploy.ovs-cni.SGd3/github.com/k8snetworkplumbingwg/ovs-cni/_kubevirtci'...
Note: switching to '2202081102-e8daf721'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at e8daf72 Add kind-1.23-vgpu provider (#746)
selecting docker as container runtime
Trying to pull quay.io/kubevirtci/gocli:2202081102-e8daf721...
Getting image source signatures
Copying blob sha256:07faf7c8ff6b82f3d29ad58234e1c82c8b9b5f770167b8ba6a987d4ecf17c3be
Copying blob sha256:07faf7c8ff6b82f3d29ad58234e1c82c8b9b5f770167b8ba6a987d4ecf17c3be
Copying config sha256:2af2351a11519a3e0610cc8e509d91a309736d5864f2cb82bb0f7517c3ad3dca
Writing manifest to image destination
Storing signatures
Error: statfs /var/run/docker.sock: no such file or directory
```

PR #233, opened 2022-07-11
Bump kubevirtci

oshoval:
Take latest changes (that affect podman socket)
See https://github.com/kubevirt/kubevirt/pull/8048

In order to able to bump kubevirtci on CNAO, we need OVS to use the same version.
Since one of the lanes of CNAO clones OVS and uses the cluster-up folder of OVS
with the cluster created by CNAO.

```release-note
None
```

### DIFF
--- a/cluster/cluster.sh
+++ b/cluster/cluster.sh
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 export KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-'k8s-1.23'}
-export KUBEVIRTCI_TAG=2205030954-99bd4d1
+export KUBEVIRTCI_TAG=2207060141-111fd50
 
 KUBEVIRTCI_REPO='https://github.com/kubevirt/kubevirtci.git'
 # The CLUSTER_PATH var is used in cluster folder and points to the _kubevirtci where the cluster is deployed from.


cluster/cluster.sh
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 export KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-'k8s-1.23'}
-export KUBEVIRTCI_TAG=2205030954-99bd4d1
+export KUBEVIRTCI_TAG=2207060141-111fd50
```
oshoval, line 16: @phoracek 
would it be fine to replace it with

`export KUBEVIRTCI_TAG=${KUBEVIRTCI_TAG:-$(curl -L -Ss https://storage.googleapis.com/kubevirt-prow/release/kubevirt/kubevirtci/latest)}`

Take the latest, unless it is already exported?
Some repos already doing that, saves time to maintain and bump,
of course risky only if kubevirtci creates breaking changes, but then we can return to pin it until we solve.

Currently since CNAO fetches ovs-cni and use some of the cluster-up folder of ovs with the CNAO created cluster, it means that every change on CNAO need to be done also on ovs else the potential mistmatch can cause problems.
If we change the code here, better to change CNAO code as well.

Another option is that CNAO once it clone the folder of ovs or any component, will clone the kubevirtci of the cluster it has in its own CNAO's repo (and also deployed)

phoracek: Interesting. I would prefer to keep it pinned though. There is not a lot of traffic on OVS CNI, so we may notice a regression/mismatch only way later after it happened. And it may be confusing to contributors to see that the code suddenly stopped working. Let's pin it, so it can regress only if somebody changes the KUBEVIRT_TAG.

phoracek:
Thanks for the bump. On it o/

PR #235, opened 2022-07-26
Bump kubevirtci

oshoval:
Release notes
```release-note
None
```


oshoval:
Thanks Petr
Can you please create a release, so it will be consumed by CNAO, and both will have the same kubevirtci version ?

phoracek:
Done

PR #237, opened 2022-08-05
Replace deprecated go get with go install

phoracek:
Without this, building of the containerized builder produces:

'go get' is no longer supported outside a module.
To build and install a command, use 'go install' with a version,
like 'go install example.com/cmd@latest'
For more information, see https://golang.org/doc/go-get-install-deprecation
or run 'go help get' or 'go help install'.

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,9 +36,7 @@ $(BASE): ; $(info  setting GOPATH...)
 
 GOLINT = $(GOBIN)/golint
 $(GOBIN)/golint: | $(BASE) ; $(info  building golint...)
-	$Q go get -u golang.org/x/lint/golint
-	# golint installation modifies go.mod and it causes golint failure so run mod tidy here
-	$(GO) mod tidy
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -36,7 +36,7 @@ $(BASE): ; $(info  setting GOPATH...)
 
 GOLINT = $(GOBIN)/golint
 $(GOBIN)/golint: | $(BASE) ; $(info  building golint...)
-	$Q go get -u golang.org/x/lint/golint
+	$Q go install -mod=mod golang.org/x/lint/golint
 	# golint installation modifies go.mod and it causes golint failure so run mod tidy here
```
rhrazdil, line 40: Do I get it correctly, that when version is specified, go.mod is not updated, and thus go mod tidy is not needed?
In that case, we could remove the comment and the go mod tidy command

phoracek: Let me try

phoracek:
LGTM analysis is acting up... Merging.

PR #240, opened 2022-08-19
Fetch Multus manifest from its repo

phoracek:
To prevent multus image and it manifest getting out of sync, fetch both
from the latest sources.

While we always fetch the latest multus image, the manifest was kept under this repo. Due to that, the two could have get out of sync and fail to deploy. This patch fixes it.

Once Multus releases stable v4 image, it would be better to pin to that explicitly, to avoid regressions introduced in their main branch.

**Release note**:

```release-note
NONE
```


maiqueb (approved):
Looks legit.

PR #241, opened 2022-08-19
Pin to OVS 3.9

phoracek:

With this patch, local development cluster will be deployed with image
and manifests of Multus 3.9 specifically. This should make sure that
they stay in sync, and that new regressions will not get to OVS CNI
unless Multus is explicitly bumped.

**Release note**:

```release-note
NONE
```


cluster/up.sh
```diff
@@ -30,5 +30,8 @@ for node in $(./cluster/kubectl.sh get nodes --no-headers | awk '{print $1}'); d
 done
 
 echo 'Deploying multus'
+curl https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/v3.9.1/deployments/multus-daemonset.yml -o cluster/multus-daemonset.yml
+MULTUS_IMAGE=ghcr.io/k8snetworkplumbingwg/multus-cni:v3.9
```
maiqueb, line 34: maybe `multus-cni:stable` ? 

They point to the same img right now, but I **assume** it moves in time. 

... tracking something *cof cof* stable.

maiqueb (approved):
Just an optional nit.

cluster/up.sh
phoracek: Issue is there is no stable branch with a manifest that would match the image, or at least I was not able to find one. So I would either have old manifest (pinning to a tag) or maybe too new one (from main). Otherwise I would commit to the trust fall!

PR #242, opened 2022-09-26
cleanup useless check in mirror test helper funcion 'CreateEmptyMirrors'

Ks89:
Quick cleanup of mirror testhelper function as requested here https://github.com/k8snetworkplumbingwg/ovs-cni/pull/227#discussion_r979462642

If createEmptyMirror executed without an error, it's safe to assume that mirror is really in ovsdb Mirror table.

### DIFF
--- a/pkg/testhelpers/mirror.go
+++ b/pkg/testhelpers/mirror.go
@@ -238,10 +238,6 @@ func CreateEmptyMirrors(bridgeName string, mirrorNames []string, ovsPortOwner st
 			_, err = addOwnerToMirror(ovsPortOwner, mirrorName)
 			gomega.Expect(err).NotTo(gomega.HaveOccurred())
 		}
-		// check existence of the empty mirrors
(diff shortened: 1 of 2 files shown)


phoracek:
@Ks89 one of the mirror tests seems flakey, would you please look into it?

```
{ Failure /tmp/src/github.com/k8snetworkplumbingwg/ovs-cni/tests/mirror_test.go:86
Expected
    <string>: 09:02:10.934759 IP 10.0.0.1 > 10.0.0.2: ICMP echo request, id 8, seq 1, length 64
    
to contain substring
    <string>: IP 10.0.0.2 > 10.0.0.1: ICMP echo reply
/tmp/src/github.com/k8snetworkplumbingwg/ovs-cni/tests/mirror_test.go:102}
```

Ks89:
We cannot reproduce this error in our environment, but it's probably related to a slower VM for kubevirtci.
We tried to increase the timeout from 10s to 30s.

phoracek:
/lgtm

Thanks!

PR #244, opened 2022-09-27
Keep all binaries in a single container image

phoracek:

Until now, each binary was kept in its own container image. That is expensive on resource and also introduces non-negligible maintenance cost.

All these containers are always deployed under a single DaemonSet and they only serve to deliver basic binaries, thus there is no reason to keep them split.

With this patch, all four container images get merged into one.

1. mirror/producer and mirror/consumer were renamed to mirror-producer and mirror-consumer respectively, to avoid need to recursively iterate all cmd/ subdirectories
2. All dockerfiles were merged and moved under a single cmd/Dockerfile
3. The three init containers deploying each of the binaries were merged into one
4. All three CNI plugins and the marker are now shipped in a single image `ovs-cni-plugin`

**Release note**:

```release-note
NONE
```


examples/ovs-cni.yml
```diff
@@ -34,33 +34,13 @@ spec:
...
-      - name: ovs-mirror-consumer
-        image: quay.io/kubevirt/ovs-cni-mirror-consumer:latest
-        command: ['cp', '/ovs-mirror-consumer', '/host/opt/cni/bin/ovs-mirror-consumer']
+        command: ["/bin/sh","-c"]
+        args:
+        args:
+          - >
+            cp /ovs /host/opt/cni/bin/ovs &&
+            cp /ovs-mirror-producer /host/opt/cni/bin/ovs-mirror-producer &&
+            cp /ovs-mirror-consumer /host/opt/cni/bin/ovs-mirror-consumer
```
qinqon, line 43: I think the binary is `ovs-cni-mirror-[producer|consumer]`

qinqon: My fault that's the image names, all good
/lgtm
/approve

PR #247, opened 2022-10-21
Use two-stage container build

phoracek:

In the current code, binaries are built locally to be only then put into the container image.

This patch aims to simplify part of that process, by containerizing builds. This should also help us avoid environment related issues, and later allow cross-platform building.

This PR also fixes incorrect .version rendering.

**Release note**:

```release-note
NONE
```


cmd/Dockerfile
```diff
@@ -1,7 +1,25 @@
+FROM quay.io/centos/centos:stream8 as builder
+
+RUN mkdir /workdir
+WORKDIR /workdir
+
+COPY go.mod .
+RUN dnf install -y golang-$(sed -En 's/^go +(.*+)$/\1/p' go.mod).*
+
+COPY . .
```
qinqon, line 9: we can just use the normal golang image and pass the tag with the version outside of this so we don't have to install

qinqon (changes requested):
Let's use the golang container since we don't need special RPMs, we can parameterize the FROM clausule to pin to the correct golang from go.mod

cmd/Dockerfile
phoracek: Issue is that the only legit pubilc golang builder I was able to find is on Docker Hub. And using that would mean facing the pull limiting...

PR #251, opened 2022-11-10
Provides a possible way to set interface's type of port.

wangyueyu64:
In the current code, only the 'internal' type which is also the defaut type of interface is available when ovs-cni create one interface for port.

  This commit added extra content to provides a possible way of setting interface's type when cni call func 'attachIfaceToBridge', In order to accommodate this function, also expanded type NetConf for ovs-cni and added extra arguments to function 'attachIfaceToBridge'.

  This will allow the user to select a custom type of interface when creating a port on bridge.

**Release note**:

```release-note

```

### DIFF
--- a/docs/cni-plugin.md
+++ b/docs/cni-plugin.md
@@ -35,6 +35,17 @@ Another example with a trunk port and jumbo frames:
 }
 ```
 
+Another example with a port which has an interface of type afxdp:
+
+```json
+{
+   "name": "overlaynet",
+   "type": "ovs",
+   "bridge": "mynet1",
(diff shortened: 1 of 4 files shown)


phoracek:
Hello @wangyueyu64, thanks for the contribution! Looks interesting.

Sorry about the broken DCO link, here is a working one: https://github.com/kubevirt/kubevirt/blob/main/CONTRIBUTING.md#contributor-compliance-with-developer-certificate-of-origin-dco

If you would have any troubles pushing the PR, I'd be happy to assist.

PR #252, opened 2022-11-10
Provides a possible way to set interface's type of one port

wangyueyu64:
In the current code, only the 'internal' type which is also the defaut type of interface is available when ovs-cni create one interface for port.
I added extra content to provides a possible way of setting interface's type when cni call func 'attachIfaceToBridge’.
In order to accommodate this function, also expanded type NetConf for ovs-cni and added extra arguments to function ‘attachIfaceToBridge'.
This will allow the user to select a custom type of interface when creating a port on bridge.

Signed-off-by: wangyueyu <wangyueyu@didiglobal.com>

**Release note**:

```release-note

```

### DIFF
--- a/docs/cni-plugin.md
+++ b/docs/cni-plugin.md
@@ -35,14 +35,14 @@ Another example with a trunk port and jumbo frames:
 }
 ```
 
-Another example with a port which has an interface of type afxdp:
+Another example with a port which has an interface of type system:
 
 ```json
 {
    "name": "overlaynet",
    "type": "ovs",
    "bridge": "mynet1",
-   "interface_type": "afxdp"
+   "interface_type": "system"
 }
 ```
 
--- a/docs/cni-plugin.md
+++ b/docs/cni-plugin.md
@@ -35,6 +35,17 @@ Another example with a trunk port and jumbo frames:
 }
 ```
 
+Another example with a port which has an interface of type afxdp:
+
+```json
+{
+   "name": "overlaynet",
+   "type": "ovs",
+   "bridge": "mynet1",
+   "interface_type": "afxdp"
+}
+```
+
 ## Network Configuration Reference
 
 * `name` (string, required): the name of the network.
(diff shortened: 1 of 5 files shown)


phoracek (changes requested):
The code looks fine. Would you please add test cases to https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/pkg/plugin/plugin_test.go covering the default type (without interface_type passed) and one for a custom type?

wangyueyu64:
> The code looks fine. Would you please add test cases to https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/pkg/plugin/plugin_test.go covering the default type (without interface_type passed) and one for a custom type?

Thank you for your reply. I will supplement the test content as soon as possible.

wangyueyu64:
> The code looks fine. Would you please add test cases to https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/pkg/plugin/plugin_test.go covering the default type (without interface_type passed) and one for a custom type?

Add new test content for interface type.

phoracek:
The test fails on the CI, probably due to afxdp support not being present in a default installation of OVS. I tried it locally and it also failed. Reading https://docs.openvswitch.org/en/latest/intro/install/afxdp/, it mentions that this specific type is experimental and compiled by default.

Could we change the type used in tests to one that is generally supported? To make sure that these tests can be executed on regular environments of other developers. `local` type may be a good choice.

phoracek:
@wangyueyu64 hello, this has been silent for a while. Can I help with anything regarding the tests?

wangyueyu64:
I am very sorry that I have replied to your message only now due to work reasons. I will use more&nbsp;general&nbsp;interface types as test cases, such as tap or system. I will submit it again after the modification is completed.



---Original---
From: "Petr ***@***.***&gt;
Date: Mon, Dec 12, 2022 20:21 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [k8snetworkplumbingwg/ovs-cni] Provides a possible way to setinterface's type of one port (PR #252)




 
@wangyueyu64 hello, this has been silent for a while. Can I help with anything regarding the tests?

phoracek:
That's all good, don't worry about that :)

wangyueyu64:
> That's all good, don't worry about that :)

wangyueyu64:
The afxdp type interface content in the test file has been replaced with the  interface of type system type.

Issue #260, opened 2023-01-10
Arm Support


MarkSpencerTan:
Hello, is OVS-CNI usable for arm nodes? I noticed that it was crosscompiled to work for the arm environment, however, I'm getting an issue where the OVS CNI pods are crashing randomly only for my arm nodes.

The only error that's visible on the pod is as follows:

```bash
ovs-cni-marker qemu: uncaught target signal 11 (Segmentation fault) - core dumped
```

Is there a way to get more verbose logging options to determine where it had a segfault?

Thanks so much!

phoracek:
Hello, could you try removing or overriding the GOARCH variable here https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/cmd/Dockerfile#L12, and then compiling? Perhaps that may help.

Is it crashing immediately on start or only after a while?

If I remember correctly, you should be able to change the log level in the manifest: https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/manifests/ovs-cni.yml.in#L63

courtland:
I opened PR #266 to at least make the plumbing for building arm64 easier. For now I am able to use this for my needs, but it would be nice to have this extended to release official binaries for both architectures.

cyclinder:
Can we release arm64 assets in https://github.com/k8snetworkplumbingwg/ovs-cni/releases？

cyclinder:
```
root@cyclinder3:~/cyclinder/ovs-cni# make GOARCH=arm64 build
/root/cyclinder/ovs-cni/build/_output/bin//go//bin//go fmt ./pkg/... ./cmd/...
/root/cyclinder/ovs-cni/build/_output/bin//go//bin//go vet ./pkg/... ./cmd/...
cd cmd/marker && /root/cyclinder/ovs-cni/build/_output/bin//go//bin//go fmt && /root/cyclinder/ovs-cni/build/_output/bin//go//bin//go vet && GOOS=linux GOARCH=amd64 CGO_ENABLED=0 GO111MODULE=on /root/cyclinder/ovs-cni/build/_output/bin//go//bin//go build -tags no_openssl -mod vendor
```

It seems `GOARCH=arm64` doesn't works...

ykulazhenkov:
full ARM support was added in https://github.com/k8snetworkplumbingwg/ovs-cni/pull/307

[...]

PR #266, opened 2023-04-14
Plumbing to optionally build for arm64

courtland:

Adds the ability to optionally build binaries for arm64 by setting `GOARCH` via Makefile variable overrides.

For example:
```bash
make GOARCH=arm64 build

make GOARCH=arm64 docker-build
```
Also fixes the non-dockerized `install-go` script to correctly download/install the versioned golang under an arm64 host.

Combined, this permits building the plugin(s) for arm64 and/or x86 on either type of host.

Related to #260 

**Release note**:
```release-note
Ability to build arm64 binaries
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ export GOROOT=$(BIN_DIR)/go/
 export GOBIN = $(GOROOT)/bin/
 export PATH := $(GOBIN):$(PATH):$(BIN_DIR)
 GOPATH = $(CURDIR)/.gopath
+GOARCH ?= amd64
 ORG_PATH = github.com/k8snetworkplumbingwg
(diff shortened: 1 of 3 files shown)


phoracek:
Hey @courtland, thanks for submitting this. The dockerized tests have failed. Would you please check it?

courtland:
It looks like the `quay.io/fedora/fedora:34-x86_64` no longer exists, I'm guessing since 34 was EOL a while back? It must have been right after I submitted this, because the tests passed for me locally then.

I will raise another issue to fix the tests and then perhaps I can rebase this PR.

phoracek:
/lgtm
/approve
/override 
pull-e2e-ovs-cni

Failed e2e on an unrelated flake

PR #279, opened 2023-09-27
check hashicorp: Add hashicorp whitelist

oshoval:
Validates that no hashicorp are used beside the whitelisted ones
(because they shifted to BSL license)

Atm no hashicorp is used, but lets protect in case someone will add.
Based on the whitelist that CNAO uses.

**Release note**:

```release-note
None
```


oshoval:
@phoracek 
can you please enable git actions here ?

oshoval:
some failures here are not ovs-mirror

oshoval:
/retest 

@phoracek
ptal, regression was fixed

PR #280, opened 2023-09-27
kubevirtci: Downgrade to 1.25

oshoval:
Align with CNAO.
Check if here the tests are passing.

Failures for example
https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/k8snetworkplumbingwg_ovs-cni/279/pull-e2e-ovs-cni/1706970379018309632
pod can't be created `failed to find plugin "calico" in path [/opt/cni/bin]`
It seems we need to update calico according https://github.com/projectcalico/calico/pull/6218
but not sure this is the problem

**Release note**:

```release-note
None
```

### DIFF
--- a/automation/check-patch.e2e.sh
+++ b/automation/check-patch.e2e.sh
@@ -12,7 +12,7 @@ teardown() {
 }
 
 main() {
-    export KUBEVIRT_PROVIDER='k8s-1.26-centos9'
+    export KUBEVIRT_PROVIDER='k8s-1.25'
 
     source automation/setup.sh
     cd ${TMP_PROJECT_PATH}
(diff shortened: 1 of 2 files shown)


oshoval:
doesnt help

maybe worth to align more stuff which are different versus how cnao runs ovs cni, such as multus

oshoval:
Note that if i tried to delete calico pods, sometimes one of them didnt manage to restart due to the same error as above
and i saw this

```
$ oc logs -n kube-system kube-apiserver-node01  | grep validated
E0927 14:28:25.716451       1 authentication.go:63] "Unable to authenticate the request" err="[invalid bearer token, service account token has been invalidated]"
```
which is also mentioned here https://github.com/projectcalico/calico/issues/5684#issuecomment-1061582582

i think the right way is to try to align everything with CNAO, to see what is the moving part that does it / update calico

PR #281, opened 2023-09-28
multus: Use thin version

oshoval:
Calico officially supports only multus thin,
thick fails to create pods [1]
We don't need here thick so lets use thin meanwhile (as cnao does).

[1] https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/k8snetworkplumbingwg_ovs-cni/279/pull-e2e-ovs-cni/1706970379018309632

It seems we are using https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/v4.0.1/deployments/multus-daemonset-thick.yml
which has `image: ghcr.io/k8snetworkplumbingwg/multus-cni:snapshot-thick`
same tag is used also on 4.0.2 and such, hence it might started to break suddenly when multus was changed.
Thin also have this problem (common tag between versions), so we should freeze by hash or so in a follow PR,
as CNAO does.

**Release note**:

```release-note
None
```

### DIFF
--- a/cluster/up.sh
+++ b/cluster/up.sh
@@ -30,5 +30,5 @@ for node in $(./cluster/kubectl.sh get nodes --no-headers | awk '{print $1}'); d
 done
 
 echo 'Deploying multus'
-./cluster/kubectl.sh apply -f https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/v4.0.1/deployments/multus-daemonset-thick.yml
+./cluster/kubectl.sh apply -f https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/v4.0.1/deployments/multus-daemonset.yml
 ./cluster/kubectl.sh -n kube-system rollout status daemonset kube-multus-ds --timeout 300s


oshoval:
/cc @phoracek 
this seems to fix the problem at least locally with some more changes (i.e k8s-1.28, latest kubevirtci, calico 3.26.1)
lets see on CI

talked with Calico people, they only test with multus thin
and say that k8s 1.25+ doesnt officially support the old calico 3.18, so worth that we consider updating it on kubevirtci

oshoval:
It passed

Note that once we bump kubevirtci on CNAO we will need to use `KUBEVIRT_WITH_MULTUS_V3` if we don't use it yet
so it will select the thin one.

phoracek:
Thanks for getting this addressed

/approve
/lgtm

phoracek:
FIY @maiqueb there seem to be compatibility issues between Calico and thick Multus

maiqueb:
> FIY @maiqueb there seem to be compatibility issues between Calico and thick Multus

@oshoval brought this to my attention yesterday.

From what I understand, it simply can't find the plugin binary in the folder where the CNIs are supposed to be ... The question is why.

maiqueb:
OK found out the reason. 

[This issue](https://github.com/k8snetworkplumbingwg/multus-cni/issues/1164) reports this same issue. It's not only calico. Basically multus now doesn't has the path to *any* CNI plugins. I don't know how this could ever have passed CI ... Oh wait... I do know: in CI it uses a different manifest 0_o 

[This PR](https://github.com/k8snetworkplumbingwg/multus-cni/pull/1161/files) broke it.

[This PR](https://github.com/k8snetworkplumbingwg/multus-cni/pull/1166) fixes it.

@oshoval where are we using this `snapshot` multus tag ?... We should **stop** using that and use a pinned version instead - `v4.0.2-thick` or `stable-thick` tags.

Can you take care of that ?

oshoval:
Nice, mentioned on PR desc
sure

PR #282, opened 2023-10-01
multus: Pin image

oshoval:
Since multus doesnt pin image correctly,
pin it before deploying multus.

See https://github.com/k8snetworkplumbingwg/multus-cni/issues/1170

**Release note**:

```release-note
None
```

### DIFF
--- a/cluster/up.sh
+++ b/cluster/up.sh
@@ -30,5 +30,8 @@ for node in $(./cluster/kubectl.sh get nodes --no-headers | awk '{print $1}'); d
 done
 
 echo 'Deploying multus'
-./cluster/kubectl.sh apply -f https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/v4.0.1/deployments/multus-daemonset.yml
+MULTUS_VERSION=v4.0.1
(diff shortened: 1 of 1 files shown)


dougbtv:
We discussed this during the maintainer's call, and... it's more complex than you might think.

* Currently we don't actually have a stable image for thick plugin yet (so that's maybe a to-do item)
* We currently like having `:snapshot` images in the main example deployment yaml, because this encourages users to deploy the latest version so that we get feedback more rapidly.
* My idea about having the git tag match the image tag is not the most straightforward release process, and its bang-to-buck ratio is low (not a lot of benefit for the effort)

Is there a reason why you can't track master for the OVS-CNI CI? Because I think that might actually be beneficial for both projects.

oshoval:
Hi
For this repo we are using thin, not thick, we don't need thick atm for this repo.
This repo doesn't run stuff a lot, so once we do, we prefer to have it stable, with whatever possible pinned and static, so we don't have surprises and can bump only when needed.

Using (as this PR does for thin plugin)
```
MULTUS_VERSION=v4.0.1
MULTUS_MANIFEST=https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/${MULTUS_VERSION}/deployments/multus-daemonset.yml
```
with $MULTUS_VERSION as the tag is legit right ?
it takes yaml of v4.0.1, and the corresponding image

Note that we never used main on the yaml link, we always pinned the version, which is another reason why to not use latest image with static yaml

PR #302, opened 2024-03-01
Fix CMDCheck for hw offload use-case

ykulazhenkov:
In CMDCheck there is a check that validates if host/cont interfaces are veth interfaces.
For HW offload this is not true. Ignore this check if ovsHWOffload is enabled
**Release note**:

```release-note
NONE
```


phoracek:
@ykulazhenkov thank you for fixing this

/retest

e0ne (approved):
LGTM

phoracek:
/approve

@ykulazhenkov thanks again for fixing this, @e0ne thanks for the review

PR #304, opened 2024-04-15
Fix problem with empty cache after reboot

ykulazhenkov:
Use `/var/lib/cni/ovs-cni/cache` dir to store config cache.
Old path `/tmp/ovscache` is cleaned up after reboot on some operating systems.
Lack of the cache entries may prevent ovs-cni to do a proper cleanup on CmdDel.
Use `/var/lib/cni/ovs-cni/cache` as a persistent cache dir.

Problem with the current code (before this patch):
1. `/tmp/ovscache` is used as a cache path by default https://github.com/k8snetworkplumbingwg/ovs-cni/blob/10d2594d0fbcce45af9c7e819ea600404100c7b8/pkg/utils/cache.go#L29
2. `/tmp` dir will be cleaned up during the host reboot on some distros
3.  CmdDel exit without an error if cache entry not found https://github.com/k8snetworkplumbingwg/ovs-cni/blob/10d2594d0fbcce45af9c7e819ea600404100c7b8/pkg/plugin/plugin.go#L485
4. `ovs-cni` will not be able to do a cleanup after the reboot, this will cause, the following state in ovs
```
d178a096-20dd-4aa1-931a-0a3e82613228
    Bridge mybr
        Port mybr
            Interface mybr
                type: internal
        Port vethd25775b1
            Interface vethd25775b1
                error: "could not open network device vethd25775b1 (No such device)"
    ovs_version: "2.17.9"
```
5. on next CmdAdd all port with an error will be removed https://github.com/k8snetworkplumbingwg/ovs-cni/blob/10d2594d0fbcce45af9c7e819ea600404100c7b8/pkg/plugin/plugin.go#L271
6. for non HW offloading use-case this will fix the issue (symptom, not the root cause)
7. for HW offloading use-case, Ports will have no error and these ports will not be released by the `cleanPorts` function. As a result VFs will stuck in the OVS and manual clean-up is required before they can be used again.

**Release note**:
```release-note
NONE
```

cc @e0ne @SchSeba

### DIFF
--- a/pkg/utils/cache.go
+++ b/pkg/utils/cache.go
@@ -19,14 +19,17 @@ package utils
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 )
 
 var (
 	// DefaultCacheDir used for caching
-	DefaultCacheDir = "/tmp/ovscache"
+	DefaultCacheDir = "/var/lib/cni/ovs-cni/cache"
+	// OldDefaultCacheDir path to the old caching dir
+	OldDefaultCacheDir = "/tmp/ovscache"
+	// used for tests
+	rootDir = ""
 )
 
 // SaveCache takes in key as string and a json encoded struct Conf and save this Conf in cache dir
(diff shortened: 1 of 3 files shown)


ykulazhenkov:
After discussion with @e0ne we decided that it is better to implement alternative option. I marked PR as draft to apply required changes.

ykulazhenkov:
> After discussion with @e0ne we decided that it is better to implement alternative option. I marked PR as draft to apply required changes.

required changes implemented. The PR is ready for reviews

PR #310, opened 2024-05-19
Add scalable functions (SF) support

naserdean:
This patch adds support for scalable functions

(https://docs.nvidia.com/networking/display/bluefielddpuosv385/scalable+functions).

They are represented on the host as auxiliary devices.

In order to for the minimal codebase changes, this patch updated the sriovnet dependency,
which already has support for SFs, together with its needed dependencies.

This patch also adds architecture detection to the Makefile.

Add support for auxiliary devices.

First time I contribute to ovs-cni.
Apologies up front for any headaches 😄 .

**Release note**:

```release-note
Support Scalable Functions
```


pkg/sriov/sriov.go
```diff
@@ -22,21 +22,34 @@ import (
...
 )
 
-// GetVFLinkName retrives interface name for given pci address
+// IsPCIDeviceName check if passed device id is a PCI device name
+func IsPCIDeviceName(deviceID string) bool {
+	return rePciDeviceName.MatchString(deviceID)
+}
+
+// IsAuxDeviceName check if passed device id is a Auxiliary device name
+func IsAuxDeviceName(deviceID string) bool {
```
adrianchiris, line 48: it might be safer to check if given device is found on the auxiliary bus like in [1]
the regex is quite loose for SFs

[1] https://github.com/containernetworking/plugins/pull/1003/files

Makefile
```diff
@@ -13,7 +13,12 @@ export GOROOT=$(BIN_DIR)/go/
 export GOBIN = $(GOROOT)/bin/
 export PATH := $(GOBIN):$(PATH):$(BIN_DIR)
 GOPATH = $(CURDIR)/.gopath
-GOARCH ?= amd64
+UNAME=$(shell uname -m | tr '[:upper:]' '[:lower:]')
```
adrianchiris, line 16: this should go in a separate PR ?

pkg/sriov/sriov.go
```diff
@@ -133,49 +161,67 @@ func getBondMembers(bond netlink.Link) ([]string, error) {
...
-	rep, err := sriovnet.GetVfRepresentor(uplink, vfIndex)
 	if err != nil {
 		return "", err
 	}
-
 	return rep, nil
 }
 
 // SetupSriovInterface moves smartVF into container namespace, rename it with ifName and also returns host interface with VF's representor device
 func SetupSriovInterface(contNetns ns.NetNS, containerID, ifName string, mtu int, deviceID string) (*current.Interface, *current.Interface, error) {
```
adrianchiris, line 201: should we rename this method as it now handles both sr-iov and auxiliary devices ?

Makefile
naserdean: oops,
indeed, will drop the change from this pr and create a new one just for this

phoracek:
Thanks for the contribution @naserdean! CC @SchSeba

Issue #331, opened 2024-11-11
Support for go uplift to 1.22.7 version and have fix for CVE-2024-24790


smoshiur1237:
We are having a vulnerability report which is related to [CVE-2024-24790](https://nvd.nist.gov/vuln/detail/CVE-2024-24790) and it has critical score. This can be fixed by uplifting go to 1.22.7. We would appreciate if you support the uplift and add the go uplift in your next official release..

phoracek:
Hi @smoshiur1237, thanks for reporting this. I have posted https://github.com/k8snetworkplumbingwg/ovs-cni/pull/332 to bump the version to the latest 1.22.9.

smoshiur1237:
Thanks @phoracek , I have left a comment and  request to update the docker file in hack folder to change the go version to 1.22.7

smoshiur1237:
@phoracek  thanks for merging the uplift. May I know when are you going to have a release to add this change? would it be in patch release or in minor release?

phoracek:
@smoshiur1237 I will issue a minor release in a second. Minor because the main branch has a new feature in it https://github.com/k8snetworkplumbingwg/ovs-cni/pull/322

phoracek:
https://github.com/k8snetworkplumbingwg/ovs-cni/releases/tag/v0.35.0 the container images should be published soon too.

I will close this issue now. @smoshiur1237 thanks for reporting it. Please feel free to reopen in case you see this CVE is not resolved in the new build.

PR #332, opened 2024-11-11
Bump go to 1.22.9

phoracek:

This is to address a CVE, reported in #331.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -108,4 +108,4 @@ replace (
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.31.0
 )
 
-go 1.22.0
+go 1.22.9
--- a/hack/docker-builder/Dockerfile
+++ b/hack/docker-builder/Dockerfile
@@ -6,7 +6,7 @@ RUN dnf -y install make git sudo gcc rsync-daemon rsync openvswitch hostname && 
 ENV GOPATH="/go"
 RUN \
     DESTINATION=/opt && \
-    VERSION=1.21.7 && \
+    VERSION=1.22.9 && \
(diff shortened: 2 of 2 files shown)


smoshiur1237:
Thanks @phoracek  for this uplift. would you please update the docker file go version in the following line to 1.22.7?
https://github.com/k8snetworkplumbingwg/ovs-cni/blob/9f9bb719fd2bd9e51aa05408eaf9781bd59d3c0e/hack/docker-builder/Dockerfile#L9

smoshiur1237:
I have the understanding that go mod file change should have minimum go version that it supports. As earlier version of 1.22.7 has the vulnerability, isn't it better to update the go version to 1.22.7 in `go mod` file and same in docker file ( `ovs-cni/hack/docker-builder/Dockerfile` ).

phoracek:
@smoshiur1237 the Go version in the docker-builder should be only relevant to the Go used in test runs. But it's a good catch - I should update it too.

The version that matters for our containerized distribution is set here https://github.com/k8snetworkplumbingwg/ovs-cni/blob/main/cmd/Dockerfile#L10, and it is based on the one listed in Go.mod. Binary builds attached to releases are currently built by me manually (no CI), but they should follow the Go.mod directive.

smoshiur1237:
@phoracek  I can understand the go version is taken from go mod file. Is there any issue to use 1.22.7 ?

phoracek:
@smoshiur1237 yes, following semver, if the fix is available in 1.22.7, it should be a part of all the newer z-stream releases.

PR #307, opened 2024-04-18
Add github actions to publish images to GH registry

ykulazhenkov:
**What this PR does / why we need it**:
Add support for multiach (linux/amd64,linux/arm64,linux/ppc64le) image build with GitHub actions and publish resulting images to Github registry.

**Special notes for your reviewer**:
Example of PR check:  https://github.com/ykulazhenkov/ovs-cni/actions/runs/8735994526
Example of push to master: https://github.com/ykulazhenkov/ovs-cni/actions/runs/8735447193
Example of push on release: https://github.com/ykulazhenkov/ovs-cni/actions/runs/8735569774

Example of artifacts: https://github.com/ykulazhenkov/ovs-cni/pkgs/container/ovs-cni

Validate binary arch:

```
root@cloud-dev-17:~/foo# docker pull --platform linux/arm64 ghcr.io/ykulazhenkov/ovs-cni:v0.0.1
v0.0.1: Pulling from ykulazhenkov/ovs-cni
0117dee8e815: Already exists 
0176767d615d: Already exists 
79424b46edd7: Already exists 
90e055d9c187: Already exists 
Digest: sha256:8b03ff2b9869dff82e07e26a80c3b9ce25c3e99c34abcc05ffae29c00c51d27c
Status: Downloaded newer image for ghcr.io/ykulazhenkov/ovs-cni:v0.0.1
ghcr.io/ykulazhenkov/ovs-cni:v0.0.1

root@cloud-dev-17:~/foo# docker create ghcr.io/ykulazhenkov/ovs-cni:v0.0.1
WARNING: The requested image's platform (linux/arm64) does not match the detected host platform (linux/amd64/v4) and no specific platform was requested
917e6995414181c0f56763b5848e58c11d46fe00932153126565b3add962d881

root@cloud-dev-17:~/foo# docker cp 917e6995414181c0f56763b5848e58c11d46fe00932153126565b3add962d881:/ovs .
Successfully copied 7.74MB to /root/foo/.

root@cloud-dev-17:~/foo# file ovs
ovs: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, Go BuildID=ve0PYRQc45bRGHwoMZ6U/6uwx9CZDgopFlyJ4crUm/IPjfOx0tRjSzN9X3qlNH/Ih4mI5QNFRHHMAWezQzy, not stripped
```

**Release note**:
```release-note
NONE
```

Based on a great work by @zeeke in another k8snetworkplumbingwg repo.

[...]

PR #336, opened 2024-11-25
Fix: multiarch build by using crosscompilation

ykulazhenkov:

Use quay.io/centos/centos:stream9 with the –platform=$BUILDPLATFORM flag,
in this case the image is pulled for the builder host's current architecture.
The BUILDOS and BUILDARCH args are used to download the right go binary for
the build platform.
Cross-compilation occurs in the builder image using TARGETOS and TARGETARCH build arguments to determine the target OS/arch.
These args are set automatically by the multiarch-build process (with docker buildx).
The final container image (registry.access.redhat.com/ubi9/ubi-minimal) is pulled for the correct target architecture, such as amd64 or arm64.

This update fixes support for multiarch builds and speeds up image building, as cross-compilation is faster than compiling on a non-native platform.

Fixes support for multiarch builds added in https://github.com/k8snetworkplumbingwg/ovs-cni/pull/307

Multiarch build was broken by https://github.com/k8snetworkplumbingwg/ovs-cni/pull/319

**Release note**:

```release-note
NONE
```


ykulazhenkov:
@phoracek @SchSeba  please take a look on this one

ykulazhenkov:
Additional context is in this message 
https://github.com/k8snetworkplumbingwg/ovs-cni/issues/260#issuecomment-2495530500

Example of the job with the code from this PR:
https://github.com/ykulazhenkov/ovs-cni/actions/runs/12008233711/job/33470460458
And artifacts
https://github.com/ykulazhenkov/ovs-cni/pkgs/container/ovs-cni-plugin/311494909?tag=v0.101.0

.github/workflows/image-build-test.yaml
```diff
@@ -27,9 +27,5 @@ jobs:
         with:
           context: .
           push: false
-          # no need to explicitly set goarch, 
-          # correct arch will be selected for each build platform 
-          build-args: |
```
ykulazhenkov, line 32: no need to set this anymore. Default for the goarch= build-arg is empty string.

cmd/Dockerfile
```diff
@@ -1,35 +1,37 @@
-FROM quay.io/centos/centos:stream9 as builder
+FROM --platform=$BUILDPLATFORM docker.io/golang:1.23 AS builder
```
oshoval, line 1: go ver is hardcoded here instead taking from go.mod
i believe we can live with it, just raising for awareness in case you think otherwsie

oshoval (commented):
Nice thanks

cmd/Dockerfile
oshoval: This doesn't affect D/S (at least our as far as i understand), and also better to usually be considered implicit
it also require changing on more than one place, which isnt what we strive to do usually,
only rarely we need a different compiler than a language level imho

ykulazhenkov: Ok, let's use go.mod if this is standard practice in kubevirt-related projects. I updated PR to rollback to explicit go download.

cmd/Dockerfile
```diff
@@ -1,27 +1,42 @@
...
-    rm go${GO_VERSION}.linux-amd64.tar.gz
+    wget https://dl.google.com/go/go${GO_VERSION}.${BUILDOS}-${BUILDARCH}.tar.gz && \
+    tar -C /usr/local -xzf go${GO_VERSION}.${BUILDOS}-${BUILDARCH}.tar.gz && \
+    rm go${GO_VERSION}.${BUILDOS}-${BUILDARCH}.tar.gz
 
-ENV PATH /usr/local/go/bin:$PATH
+ENV PATH=/usr/local/go/bin:$PATH
 
-COPY . .
+RUN go mod download
```
oshoval, line 37: curious, why we need this suddenly ?
aint

ykulazhenkov: First, we copy `go.mod` and `go.sum` to download dependencies explicitly, then we copy other code files. This optimizes image rebuilds when only source code changes, allowing Docker to use cached dependencies.
This construction is employed to prevent re-downloading the dependencies whenever possible.
This is pretty common practice. For example, kubebuilder will generate Dockerfile that use such optimization https://github.com/kubernetes-sigs/kubebuilder/blob/cbc6e383c342f1337ab37ee4aa0755957a01f9c7/pkg/plugins/golang/v4/scaffolds/internal/templates/dockerfile.go#L46

ashokpariya0:
@ykulazhenkov We should update the GOARCH ?= amd64 line (as seen in [this Makefile](https://github.com/k8snetworkplumbingwg/ovs-cni/blob/546b35151345abaa92cf969aae215ba1aafd763f/Makefile#L16)) to:
GOARCH ?= $(shell uname -m | sed 's/x86_64/amd64/')
This change will dynamically set GOARCH based on the host architecture, converting x86_64 to amd64 as needed.

@oshoval changes look good to me, assuming the idea is to provide multiplatform support only through GitHub Actions, exclusively for the Docker container runtime, and not through make commands or for the Podman container runtime.

oshoval:
> @oshoval changes look good to me, assuming the idea is to provide multiplatform support only through GitHub Actions, exclusively for the Docker container runtime, and not through make commands or for the Podman container runtime.

thanks

at the end we will do need also images that are multi platform though isnt it? (as the issue / PRs you work on)
it is fine to do so in steps, as long as those steps can co exists

ashokpariya0:
> > @oshoval changes look good to me, assuming the idea is to provide multiplatform support only through GitHub Actions, exclusively for the Docker container runtime, and not through make commands or for the Podman container runtime.
> 
> thanks
> 
> at the end we will do need also images that are multi platform though isnt it? (as the issue / PRs you work on) it is fine to do so in steps, as long as those steps can co exists

Yes, we can consider adding multiplatform support using a Makefile(using make commands) for both Docker and Podman in the future. This would allow us to build images on our local machine, in a specific environment, or within any CI system, giving us more control over the build process and environment.
For Podman, there are no additional dependencies required to build multiplatform images, while for Docker, we would need the docker buildx tool.

ykulazhenkov:
> We should update the GOARCH ?= amd64 line (as seen in [this Makefile](https://github.com/k8snetworkplumbingwg/ovs-cni/blob/546b35151345abaa92cf969aae215ba1aafd763f/Makefile#L16)) to:
GOARCH ?= $(shell uname -m | sed 's/x86_64/amd64/')
This change will dynamically set GOARCH based on the host architecture, converting x86_64 to amd64 as needed.

done

oshoval (approved):
Thanks, very nice

PR #311, opened 2024-05-21
Add PodUID to the OVSDB

jiayoukun:
**What this PR does / why we need it**:

By adding the UID of a Pod to the external_ids field of the corresponding veth interface in the OVSDB, it becomes convenient for the Pod object to query the OVSDB using the UID to find the vethName of the Pod on the host machine, and then deliver OVS flow tables accordingly.

Currently, the host machine's veth interface in the OVSDB only stores the Netns field, which is useful, but for Pod objects, their netns is not publicly accessible. This change addresses this limitation by allowing Pod objects to query the OVSDB directly using their UID.

**Special notes for your reviewer**:

**Release note**:

```release-note
Added the Pod UID to the external_ids field of the corresponding veth interface in the OVSDB.
This change allows querying the vethName of a Pod on the host machine using its UID, improving the ability to deliver OVS flow tables for Pod objects.
```

PR #339, opened 2024-12-05
Add port pod namespace name to port creation

filanov:
Adding external_id to port creation with pod namespace/name it will enable easy serch between ovs port and port inside the container

In case another SDN controller need want to use ovs ports in oder to add specific flows related to the network configuration inside the pod it will be easer to find the relevant port in OVS


filanov:
Looks like it's the same as https://github.com/k8snetworkplumbingwg/ovs-cni/pull/311 
Looking at the code i think that having the ID is even better, closing this PR

PR #347, opened 2025-01-28
feat: simplify renovate config

ksimon1:
feat: simplify renovate config

Disable regular updates, because they did not work as expected, only security updates are kept. Removed configuration for go version, since it did not work as we need. Removed grouping of the dependency PRs, since it complicated the bumping of the dependencies

**Release note**:
```
NONE
```

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -9,8 +9,6 @@
     "release-0.34"
   ],
   "prConcurrentLimit": 3,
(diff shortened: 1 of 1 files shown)


phoracek (approved):
/lgtm
/approve

phoracek:
Thanks @ksimon1! Much appreciated

PR #427, opened 2026-02-02
Support vhost_vdpa devices

bgartzi:
This PR adds support for vhost_vdpa devices to ovs-cni.

Most of the steps needed to set up a vhost_vdpa device are similar to the ones of sriov devices. However, instead of setting the mac address to the VF that is moved into the pod namespace, it is the vdpa device that needs to be updated.

Setting a mac address to vdpa devices is supported from kernel 6.12 on.

If I understand correctly, there's no proper way to make IPAM work directly as the vdpa interface would be under the VM, not the host or the pod.

```release-note
Support vhost-vdpa devices for clusters running on kernels newer than 6.12
```

### DIFF
--- a/docs/vdpa.md
+++ b/docs/vdpa.md
@@ -1,0 +1,134 @@
+# VDPA
+
+This document covers ovs-cni support of vdpa devices. It describes how
+they are supported, how to configure the environment and how to attach
+vdpa devices to a ovs bridge by using ovs-cni.
+
+## Support
+
+VDPA devices can be backed up by a SR-IOV virtual function or a VDUSE
+management device.
+
+VDPA devices can be bound to a virtio or a vhost driver.
+
+Currently, ovs-cni supports VDPA devices backed up by SR-IOV virtual
+functions and bound to the vhost_vdpa driver.
+
+## Environment preparation
+
+There are several steps to take before attaching a vdpa device to a pod
+through ovs-cni
+
+### Prerequisites
+
+- Mellanox ConnectX-6 dx NIC
+- [sriov-network-device-plugin](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin)
+- [multus-cni](https://github.com/k8snetworkplumbingwg/multus-cni)
+- [network-resources-injector](https://github.com/k8snetworkplumbingwg/network-resources-injector)
+
+### VDPA device creation
+
+VDPA devices can be created by using the
+[sriov-network-operator][sriov-net-op]. However, this document covers
+creating vdpa devices without relying on the operator.
+
+#### ConnectX-6 Dx
+
+Create some VFs by:
+
+```bash
+echo 2 > /sys/class/net/ens1f0np0/device/sriov_numvfs
+```
+
+Unbind the mlx5 driver, change the e-switch mode into switchdev, and
+bind the drivers back:
+
+```bash
+echo 0000:65:00.0 > sys/bus/pci/drivers/mlx5_core/unbind
+echo 0000:65:00.1 > sys/bus/pci/drivers/mlx5_core/unbind
+devlink dev eswitch set pci/0000:65:00.0 mode switchdev
+devlink dev eswitch set pci/0000:65:00.1 mode switchdev
+echo 0000:65:00.0 > sys/bus/pci/drivers/mlx5_core/bind
+echo 0000:65:00.1 > sys/bus/pci/drivers/mlx5_core/bind
+```
+
+Load drivers
+
+```bash
+modprobe vdpa
+modprobe vhost_vdpa
+modprobe mlx5_vdpa
+```
+
+Check the available mgmtdevs:
+
+```bash
+vdpa mgmtdev show
+```
+
+Create the vdpa devices:
+```bash
+vdpa dev add name vdpa:0000:65:00.0 mgmtdev pci/0000:65:00.0 mac 02:11:22:33:44:55
+vdpa dev add name vdpa:0000:65:00.1 mgmtdev pci/0000:65:00.1 mac 02:11:22:33:44:66
+```
+
+### Device plugin configuration
+
+Just for demo purposes this document proposes a device plugin configuration
+that selects devices based on the PF name. Write/edit
+`/etc/pcidp/config.json` so it looks like:
+
+```json
+{
+    "resourceList": [{
+            "resourceName": "mlx-vdpa",
+            "selectors": {
+                "pfNames": ["ens1f0np0"]
+            }
+        }
+    ]
+}
+```
+
+Now the device plugin should advertise the vdpa devices created in the
+previous section.
+
+## NAD configuration
+
+ovs-cni relies on the device-info file to know whether the underlying
+device being configured for the network is a vdpa device. To make the
+CNI device-info aware, the path to the DeviceInfo file must be passed by
+multus. To do so, the `CNIDeviceInfoFile` capability must be enabled in
+the ovs network.
+
+Assuming that an OVS bridge called `br-vdpa` already exists, a network
+attachment definition must be created now, glueing everything together:
+```yaml
+apiVersion: k8s.cni.cncf.io/v1
+kind: NetworkAttachmentDefinition
+metadata:
+  annotations:
+    k8s.v1.cni.cncf.io/resourceName: openshift.io/mlx-vdpa
+  name: ovs-vdpa-net
+spec:
+  config: |-
+    {
+        "cniVersion": "4.0.0",
+        "name": "ovs-vdpa-net",
+        "type": "ovs",
+        "capabilities": {
+            "CNIDeviceInfoFile": true
+        },
+        "bridge": "br-vdpa",
+        "ipam": {}
+    }
+```
+
+Remember that ipam is not supported for `vhost_vdpa`, as there is not
+interface in the host or pod that can be assigned an IP directly. It
+must be the guest doing that, which the CNI does not have access to.
+
+Pods attached to `ovs-vdpa-net` will be configured a `vhost_vdpa` device
+that is attached to the `br-vdpa` OVS bridge. The ovs-cni will configure
+the network device and bridge port attachment. The network resources
+injector will handle the resource request injection into the pod spec.
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -17,8 +17,10 @@ package common
 import (
 	"errors"
 	"fmt"
+	"log"
 	"sort"
 
+	"github.com/k8snetworkplumbingwg/ovs-cni/pkg/ovsdb"
 	"github.com/k8snetworkplumbingwg/ovs-cni/pkg/types"
 )
 
@@ -112,4 +114,21 @@ func GetBridgeName(bridgeName, ovnPort string) (string, error) {
 	}
 
 	return "", fmt.Errorf("failed to get bridge name")
+}
+
+// CleanPorts removes all ports whose interfaces have an error.
+func CleanPorts(ovsDriver *ovsdb.OvsBridgeDriver) error {
+	ifaces, err := ovsDriver.FindInterfacesWithError()
+	if err != nil {
+		return fmt.Errorf("clean ports: %v", err)
+	}
+	for _, iface := range ifaces {
+		log.Printf("Info: interface %s has error: removing corresponding port", iface)
+		if err := ovsDriver.DeletePort(iface); err != nil {
+			// Don't return an error here, just log its occurrence.
+			// Something else may have removed the port already.
+			log.Printf("Error: %v\n", err)
+		}
+	}
+	return nil
 }
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -29,6 +29,7 @@ import (
 	"github.com/containernetworking/cni/pkg/skel"
 	cnitypes "github.com/containernetworking/cni/pkg/types"
 	current "github.com/containernetworking/cni/pkg/types/100"
+	"github.com/containernetworking/cni/pkg/version"
 	"github.com/containernetworking/plugins/pkg/ipam"
 	"github.com/containernetworking/plugins/pkg/ns"
 
@@ -496,4 +497,19 @@ func CacheLoadAndCheck(args *skel.CmdArgs, netconf *types.NetConf) (*types.Cache
 	}
 
 	return cache, nil
+}
+
+// ParsePrevResult parses the previous CNI result from netconf
+func ParsePrevResult(netconf *types.NetConf) (*current.Result, error) {
+	if netconf.NetConf.RawPrevResult == nil {
+		return nil, fmt.Errorf("Required prevResult missing")
+	}
+	if err := version.ParsePrevResult(&netconf.NetConf); err != nil {
+		return nil, err
+	}
+	result, err := current.NewResultFromResult(netconf.NetConf.PrevResult)
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
 }
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -20,6 +20,10 @@ import (
 	"log"
 	"sort"
 
+	"github.com/vishvananda/netlink"
+
+	current "github.com/containernetworking/cni/pkg/types/100"
+
 	"github.com/k8snetworkplumbingwg/ovs-cni/pkg/ovsdb"
 	"github.com/k8snetworkplumbingwg/ovs-cni/pkg/types"
 )
@@ -130,5 +134,14 @@ func CleanPorts(ovsDriver *ovsdb.OvsBridgeDriver) error {
 			log.Printf("Error: %v\n", err)
 		}
 	}
+	return nil
+}
+
+func RefetchIface(iface *current.Interface) error {
+	link, err := netlink.LinkByName(iface.Name)
+	if err != nil {
+		return fmt.Errorf("failed to refetch interface %s: %v", iface.Name, err)
+	}
+	iface.Mac = link.Attrs().HardwareAddr.String()
 	return nil
 }
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -30,6 +30,7 @@ import (
 	cnitypes "github.com/containernetworking/cni/pkg/types"
 	current "github.com/containernetworking/cni/pkg/types/100"
 	"github.com/containernetworking/cni/pkg/version"
+	"github.com/containernetworking/plugins/pkg/ip"
 	"github.com/containernetworking/plugins/pkg/ipam"
 	"github.com/containernetworking/plugins/pkg/ns"
 
@@ -537,4 +538,30 @@ func ExtractInterfaces(args *skel.CmdArgs, result *current.Result, ovsHWOffloadE
 	}
 
 	return hostIntf, contIntf, nil
+}
+
+// ValidateNetnsAttachment validates IPs and routes in the container namespace
+func ValidateNetnsAttachment(args *skel.CmdArgs, result *current.Result, contIntf current.Interface, ovsHWOffloadEnable bool) error {
+	netns, err := ns.GetNS(args.Netns)
+	if err != nil {
+		return fmt.Errorf("failed to open netns %q: %v", args.Netns, err)
+	}
+	defer func() { _ = netns.Close() }()
+
+	// Check prevResults for ips and routes against values found in the container
+	return netns.Do(func(_ ns.NetNS) error {
+		// Check interface against values found in the container
+		if err := ValidateInterface(contIntf, false, ovsHWOffloadEnable); err != nil {
+			return err
+		}
+
+		if err := ip.ValidateExpectedInterfaceIPs(args.IfName, result.IPs); err != nil {
+			return err
+		}
+
+		if err := ip.ValidateExpectedRoute(result.Routes); err != nil {
+			return err
+		}
+		return nil
+	})
 }
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -21,10 +21,16 @@ import (
 	"log"
 	"net"
 	"sort"
+	"time"
 
+	"github.com/j-keck/arping"
 	"github.com/vishvananda/netlink"
 
+	"github.com/containernetworking/cni/pkg/skel"
+	cnitypes "github.com/containernetworking/cni/pkg/types"
 	current "github.com/containernetworking/cni/pkg/types/100"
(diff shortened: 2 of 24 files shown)


bgartzi:
Hi @phoracek, confirmed that in case we are dividing the vdpa logic into a separate bin we would need to extend the `OVSNetwork` CRD in the sriov-network-operator source.

Apart from further development being needed, this could be error prone, as we would need users to indicate that we are dealing with vdpa devices both in the `SriovNetworkNodePolicySpec`'s `VdpaType` field, plus on a new field of `OVSNetwork`.

If maintainership and maintenance are still a concern from an ovs-cni point of view, would in any way isolating plugin's commands (add,del...) from each target device type help in any sort of way? 

If you still prefer having logic separated in two separate bins, I can work on that.

phoracek:
Just for the record - we met with Benat offline and agreed that we will keep this new functionality behind `ovs` binary, splitting the VDPA logic from the core internally.

bgartzi:
Which I will soon provide updates for. Thanks @phoracek

bgartzi:
Now pkg/plugin/plugin is refactored into backend plugin implementations into which the main plugin/plugin.go implementation dispatches early during the plugin execution.

This got rather big due to the refactoring. We might consider splitting it into different PRs if that somehow helps to the review.

Comparing to the last version I submitted, this time the vdpa detection relies on the device info file. That way we don't try to reach the vdpa netlink as soon as the plugin starts to run to know whether a vdpa device is using the device set by deviceID as the mgmtdev. Instead, we read the deviceinfo file and based on its type we decide what to do. This could make it easier in the future if we are also covering vduse devies. However this requires networks using vdpa devices to enable the `CNIDeviceInfoFile` capability in the NAD's capabilities.

PS I've just noticed this needs a rebase :/ will be back tomorrow.

phoracek:
@bgartzi thanks for the refactoring. Please address the Gemini review, either implement fixes, or reply in thread why are you not implementing them, and then resolve the threads.

pkg/common/common.go
```diff
@@ -0,0 +1,594 @@
...
+			for v := minID; v <= maxID; v++ {
+				vlans[v] = true
+			}
+		}
+		var id uint
+		if item.ID != nil {
+			id = *item.ID
+			if minID > 4096 {
+				return nil, errors.New("incorrect trunk id parameter")
+			}
```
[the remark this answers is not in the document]
bgartzi: This was carried over from previous code and is not specific to this PR. It will be addressed by https://github.com/k8snetworkplumbingwg/ovs-cni/issues/511

pkg/common/common.go
```diff
@@ -0,0 +1,594 @@
...
+
+func SplitVlanIds(trunks []*types.Trunk) ([]uint, error) {
+	vlans := make(map[uint]bool)
+	for _, item := range trunks {
+		var minID uint = 0
+		var maxID uint = 0
+		if item.MinID != nil {
+			minID = *item.MinID
+			if minID > 4096 {
+				return nil, errors.New("incorrect trunk minID parameter")
```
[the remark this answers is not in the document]
bgartzi: This was carried over from previous code and is not specific to this PR. It will be addressed by https://github.com/k8snetworkplumbingwg/ovs-cni/issues/510

pkg/common/common.go
```diff
@@ -0,0 +1,594 @@
...
+			}
+		}
+		if item.MaxID != nil {
+			maxID = *item.MaxID
+			if maxID > 4096 {
+				return nil, errors.New("incorrect trunk maxID parameter")
```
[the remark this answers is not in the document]
bgartzi: This was carried over from previous code and is not specific to this PR. It will be addressed by https://github.com/k8snetworkplumbingwg/ovs-cni/issues/510

pkg/vdpa/vdpa.go
```diff
@@ -0,0 +1,253 @@
...
+		}
+		contIface.Mac = vdpaMacAddr.String()
+	}
+
+	if mtu != 0 {
+		if err = netlink.LinkSetMTU(repLink, mtu); err != nil {
+			return nil, nil, err
+		}
+
+		vfNetName, err := sriov.GetNetVF(deviceID)
```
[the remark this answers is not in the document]
bgartzi: Just because that is the only vdpa device type that is supported by this implementation. VDUSE is not supported yet, which would be supported by a devicetype such as `VdpaDeviceTypeUserVhost` instead of the current `VdpaDeviceTypeKernelVhost`.

pkg/vdpa/vdpa.go
```diff
@@ -0,0 +1,253 @@
...
+func getVdpaDeviceFromID(deviceID string) (*kvdpa.VdpaDevice, error) {
+	// Don't panic if the device is not vdpa
+	if deviceID == "" {
+		return nil, nil
+	}
+
+	var vdpaDev *kvdpa.VdpaDevice
+	vdpaDevs, err := kvdpa.GetVdpaDevicesByPciAddress(pciAddressFrom(deviceID))
+	if err != nil {
+		return nil, fmt.Errorf("failed to run vdpa netlink command")
```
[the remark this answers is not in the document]
bgartzi: Duplicate of https://github.com/k8snetworkplumbingwg/ovs-cni/pull/427#discussion_r2754246033

pkg/vdpa/vdpa.go
```diff
@@ -0,0 +1,180 @@
...
+	}
```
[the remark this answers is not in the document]
bgartzi: addressed on the last pushed version.

pkg/vdpa/vdpa.go
```diff
@@ -0,0 +1,180 @@
...
+	if vdpaDev == nil {
+		return VdpaDeviceTypeNone, nil
+	}
+
+	switch (*vdpaDev).Driver() {
+	case kvdpa.VhostVdpaDriver:
+		return VdpaDeviceTypeKernelVhost, nil
+	default:
+		return VdpaDeviceTypeNone, fmt.Errorf("unknown vdpa device type")
+	}
```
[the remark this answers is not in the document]
bgartzi: Addressed on the last pushed version.

bgartzi:
@phoracek I addressed the comments on the details shared in error messages. Other comments were related to issues that existed on the code before the refactoring, so I opened a couple of issues to address them later if applicable: #510, #511.

docs/vdpa.md
```diff
@@ -0,0 +1,134 @@
...
+the ovs network.
+
+Assuming that an OVS bridge called `br-vdpa` already exists, a network
+attachment definition must be created now, glueing everything together:
+```yaml
+apiVersion: k8s.cni.cncf.io/v1
+kind: NetworkAttachmentDefinition
+metadata:
+  annotations:
+    k8s.v1.cni.cncf.io/resourceName: openshift.io/cx6dxsriovvdpa
```
phoracek: Please make sure that the examples are consistent - should this be referencing the "mlx-vdpa" resource defined above?

phoracek:
## Review

Nice work splitting the monolithic `plugin.go` into per-backend packages — the architecture direction is good. I found a few issues, mostly around the refactoring introducing subtle behavioral changes compared to `origin/main`, plus a couple of bugs in the new vdpa code.

### Bugs

**1. Nil pointer dereference in `validateKernelVdpaVhost`** — `pkg/vdpa/vdpa.go:203-218`

`getVdpaDeviceFromID` returns `nil, nil` when no vdpa devices are found for the PCI address (`len(vdpaDevs) == 0`). `validateKernelVdpaVhost` checks `err != nil` but not `vdpaDev == nil`, then dereferences via `getVdpaMacAddr(vdpaDev)` at line 210 and `(*vdpaDev).Name()` at line 218 — this will panic if the vdpa device disappears between ADD and CHECK.

The root cause is `getVdpaDeviceFromID` (line 46-65): when `deviceID` is non-empty but 0 vdpa devices are found, it silently returns `nil, nil` instead of an explicit error. This also affects CmdAdd — the error surfaces as the misleading "non-vdpa devices can not be configured as such" instead of "no vdpa device found for PCI address X".

Fix — return an error for the 0-device case in `getVdpaDeviceFromID`:
```go
if len(vdpaDevs) == 0 {
    return nil, fmt.Errorf("no vdpa device found for PCI address %q", deviceID)
}
```

**2. Error swallowed in vdpa CmdAdd cleanup** — `pkg/vdpa/plugin.go:117-126`

When `AttachIfaceToBridge` fails, the original error is overwritten by the cleanup result. If cleanup succeeds, the function returns `nil` — silently swallowing the real failure:

```go
if err = common.AttachIfaceToBridge(...); err != nil {
    _, _, err = common.CleanupOvsPortBestEffort(...)  // overwrites err
    if err != nil {
        log.Printf("Failed best-effort cleanup: %v", err)
    }
    return err  // returns nil if cleanup succeeded
}
```

Fix — save the original error before cleanup and return it. Note: the same error-shadowing pattern exists in `pkg/sriov/plugin.go:122-131` and `pkg/veth/plugin.go:95-101` (the `defer` reassigns `err` from cleanup, masking the original error if cleanup succeeds).

### Behavioral regressions from the refactoring

**3. `CmdDel` cache cleanup defer always fires, even on failure** — `pkg/plugin/plugin.go:92-108`

The cache cleanup defer on line 93 checks `err == nil`, but `err` is from `LoadConfFromCache` (line 80) and is always nil at this point — the `return veth.CmdDel(...)` / `return sriov.CmdDel(...)` statements don't assign to `err`, so the cache is unconditionally deleted even when the sub-command fails. In the original code all CmdDel logic was inline on the same `err` variable, so a failing `ipam.ExecDel` or `ReleaseVF` would leave `err` non-nil and preserve the cache for kubelet to retry.

Fix: `err = veth.CmdDel(args, cache); return err` (and same for the other two branches).

**4. Missing `refetchIface(hostIface)` after bridge attachment** — `pkg/veth/plugin.go` (after line 100), `pkg/sriov/plugin.go` (after line 121)

The original had this at `plugin.go:366-369` to update the host MAC after OVS may change it; without it the CNI result can contain a stale MAC.

**5. `veth.CmdDel` discards `DelLinkByName` errors** — `pkg/veth/plugin.go:185`

This discards the `err` from `DelLinkByName` — the original at `plugin.go:663-667` used `if err := cleanPorts(...); return err` to preserve the outer `err`, so non-NSPath/LinkNotFound errors were still returned.

### Design nits

**6. Asymmetric dispatch between CmdAdd and CmdDel** — `pkg/plugin/plugin.go`

The dispatch order is inconsistent:
- CmdAdd (line 64-72): vdpa → veth → sriov
- CmdDel (line 100-108): veth → vdpa → sriov
- CmdCheck (line 120-133): veth → vdpa → sriov

Currently functionally equivalent because vdpa always has `DeviceID` set (so it won't match the veth path). But the inconsistency is fragile — the dispatch order should be consistent across all three operations.

**7. `ovsPortConfig` unexported struct returned by exported function** — `pkg/common/common.go:47`

`type ovsPortConfig struct` (lowercase) is returned by exported `func OvsPortConfig(...)`. Callers can't name the type, declare variables of it, or pass it by name. Consider exporting as `OvsPortConfig`.

phoracek (changes requested):

pkg/vdpa/vdpa.go
```diff
@@ -0,0 +1,253 @@
...
+		return nil, fmt.Errorf("failed to get devices for PCI address %q: %w", deviceID, err)
+	}
+
+	if len(vdpaDevs) == 1 {
+		vdpaDev = &vdpaDevs[0]
+	} else if len(vdpaDevs) > 1 {
+		return nil, fmt.Errorf("multiple vdpa devices attached to the same pci mgmt device are not supported")
+	}
+
+	return vdpaDev, nil
```
phoracek, line 66: **Bug: nil pointer dereference in `validateKernelVdpaVhost`**

When `len(vdpaDevs) == 0`, this returns `nil, nil`. Downstream in `validateKernelVdpaVhost` (line 210), the nil `vdpaDev` is passed to `getVdpaMacAddr` which dereferences `(*vdpaDevice).Name()` — panic.

This also affects `CmdAdd`: if `IsVdpa(deviceInfo)` is true but no vdpa device is found for the PCI address, the error surfaces as the misleading "non-vdpa devices can not be configured as such" instead of a clear message.

Fix — return an error for the 0-device case:
```go
if len(vdpaDevs) == 0 {
    return nil, fmt.Errorf("no vdpa device found for PCI address %q", deviceID)
}
```

pkg/vdpa/plugin.go
```diff
@@ -0,0 +1,230 @@
...
+		portCfg.Trunks,
+		portCfg.Type,
+		netconf.InterfaceType,
+		args.Netns,
+		ovnPort,
+		contPodUid,
+	); err != nil {
+		// Unlike veth pair, OVS port will not be automatically removed
+		// if the following IPAM configuration fails and netns gets removed.
+		_, _, err = common.CleanupOvsPortBestEffort(ovsBridgeDriver, args.IfName, args.Netns)
```
phoracek: **Bug: original error swallowed by cleanup**

When `AttachIfaceToBridge` fails, the original `err` is overwritten here by the cleanup result. If cleanup succeeds (returns `nil`), `return err` on line 125 returns `nil` — silently swallowing the bridge-attach failure.

Fix — save the original error:
```go
origErr := err
_, _, cleanupErr := common.CleanupOvsPortBestEffort(ovsBridgeDriver, args.IfName, args.Netns)
if cleanupErr != nil {
    log.Printf("Failed best-effort cleanup: %v", cleanupErr)
}
return origErr
```

Note: the same error-shadowing pattern exists in `pkg/sriov/plugin.go:126` and `pkg/veth/plugin.go:104` — the `defer` reassigns `err` from cleanup, masking the original error if cleanup succeeds.

pkg/plugin/plugin.go
```diff
@@ -555,116 +97,15 @@ func CmdDel(args *skel.CmdArgs) error {
...
-	bridgeName, err := getBridgeName(ovsDriver, cache.Netconf.BrName, ovnPort, cache.Netconf.DeviceID)
-	if err != nil {
-		return err
-	}
-
-	ovsBridgeDriver, err := ovsdb.NewOvsBridgeDriver(bridgeName, cache.Netconf.SocketFile)
-	if err != nil {
-		return err
+	if !common.IsOvsHardwareOffloadEnabled(cache.Netconf.DeviceID) {
+		return veth.CmdDel(args, cache)
```
phoracek: **Regression: cache cleanup defer always fires, even on failure**

The defer on line 92 checks `err == nil`, but `err` is from `LoadConfFromCache` (line 80) — it's always `nil` here. These `return veth.CmdDel(...)` / `return vdpa.CmdDel(...)` / `return sriov.CmdDel(...)` statements don't assign to `err`, so the cache is unconditionally deleted even when the sub-command fails.

In the original code, all CmdDel logic was inline on the same `err` variable, so a failing `ipam.ExecDel` or `ReleaseVF` would leave `err` non-nil and preserve the cache for kubelet to retry.

Fix:
```go
err = veth.CmdDel(args, cache)
return err
```
(and same for the other two branches)

pkg/plugin/plugin.go
```diff
@@ -555,116 +97,15 @@ func CmdDel(args *skel.CmdArgs) error {
...
-	}
-	bridgeName, err := getBridgeName(ovsDriver, cache.Netconf.BrName, ovnPort, cache.Netconf.DeviceID)
-	if err != nil {
-		return err
-	}
-
-	ovsBridgeDriver, err := ovsdb.NewOvsBridgeDriver(bridgeName, cache.Netconf.SocketFile)
-	if err != nil {
-		return err
+	if !common.IsOvsHardwareOffloadEnabled(cache.Netconf.DeviceID) {
```
phoracek: **Nit: asymmetric dispatch order**

CmdAdd dispatches: vdpa → veth → sriov (line 64-72)
CmdDel dispatches: veth → vdpa → sriov (here)
CmdCheck dispatches: veth → vdpa → sriov (line 120-133)

Currently functionally equivalent because vdpa always has `DeviceID` set. But the inconsistency is fragile — the dispatch order should be consistent across all three operations.

pkg/veth/plugin.go
```diff
@@ -0,0 +1,219 @@
...
+		portCfg.VlanTag,
+		portCfg.Trunks,
+		portCfg.Type,
+		netconf.InterfaceType,
+		args.Netns,
+		ovnPort,
+		contPodUid,
+	); err != nil {
+		return err
+	}
```
phoracek, line 100: **Regression: missing `refetchIface(hostIface)` after bridge attachment**

The original code had `refetchIface(hostIface)` at `plugin.go:366-369` to update the host MAC after OVS may change it. Without it, the CNI result (line 111-113) can contain a stale MAC.

Same issue in `pkg/sriov/plugin.go` after line 121.

pkg/veth/plugin.go
```diff
@@ -0,0 +1,219 @@
...
+		if portFound {
+			if err := ip.DelLinkByName(portName); err != nil {
+				log.Printf("Failed best-effort cleanup of %s: %v", portName, err)
+			}
+		}
+		return nil
+	}
+
+	// removes all ports whose interfaces have an error
+	return common.CleanPorts(ovsBridgeDriver)
```
phoracek: **Regression: `DelLinkByName` errors discarded**

When `DelLinkByName` (line 170) returns a non-NSPath/non-LinkNotFound error, that `err` is silently discarded here — `CleanPorts` overwrites it with its own result.

The original at `plugin.go:663-667` preserved the outer `err` with:
```go
if err := cleanPorts(ovsBridgeDriver); err != nil {
    return err
}
return err  // returns the original DelLinkByName error
```

pkg/common/common.go
```diff
@@ -0,0 +1,594 @@
...
+	"github.com/k8snetworkplumbingwg/ovs-cni/pkg/ovsdb"
+	"github.com/k8snetworkplumbingwg/ovs-cni/pkg/types"
+)
+
+const (
+	portTypeAccess = "access"
+	portTypeTrunk  = "trunk"
+)
+
+type ovsPortConfig struct {
```
phoracek: **Nit: unexported struct returned by exported function**

`ovsPortConfig` (lowercase) is returned by exported `OvsPortConfig(...)`. Callers in `sriov`, `vdpa`, and `veth` packages can access fields but can't name the type, declare variables of it, or pass it by name. Consider exporting as `OvsPortConfig`.

phoracek (commented):
Nice work splitting the monolithic `plugin.go` into per-backend packages — the architecture direction is good. I found a few issues, mostly around the refactoring introducing subtle behavioral changes compared to `origin/main`, plus a couple of bugs in the new vdpa code. See inline comments.

phoracek:
Hey @bgartzi just couple of smaller issues, but overall this looks good to me. Thanks a lot for putting the effort into this - it looks much better than the current origin/main code hehe

docs/vdpa.md
bgartzi: yes, it should: my bad.

pkg/plugin/plugin.go
bgartzi: I moved vdpa onto veth in del and check.

PS I mean as soon as I force push a new version.

bgartzi:
Thanks for the review @phoracek, good finds those. I addressed them in the last forced push :)

pkg/veth/plugin.go
```diff
@@ -0,0 +1,229 @@
...
+		return err
+	}
+
+	// Refetch the host interface MAC since OVS may change it when
+	// attaching the port to the bridge.
+	if err = common.RefetchIface(hostIface); err != nil {
+		return err
+	}
+
+	defer func() {
```
phoracek, line 102: The deferred OVS port cleanup is registered *after* `AttachIfaceToBridge` returns successfully. If `AttachIfaceToBridge` fails partway through — e.g., `CreatePort` (common.go:179) succeeds but `LinkSetUp` (common.go:191) fails — the function returns the error at line 99 before the `defer` at line 108 is registered, so the orphaned OVS port is never cleaned up.

The same issue exists in `sriov/plugin.go` (line 129). Interestingly, `vdpa/plugin.go` (lines 119-125) handles this correctly with inline cleanup on `AttachIfaceToBridge` failure.

Consider either moving the defer before the `AttachIfaceToBridge` call, or adding inline cleanup in the error path like the vdpa plugin does.

pkg/vdpa/plugin.go
```diff
@@ -0,0 +1,230 @@
...
+		return err
+	}
+
+	contNetns, err := ns.GetNS(args.Netns)
+	if err != nil {
+		return fmt.Errorf("failed to open netns %q: %v", args.Netns, err)
+	}
+	defer func() { _ = contNetns.Close() }()
+
+	origIfName, err := sriov.GetVFLinkName(netconf.DeviceID)
```
phoracek: `sriov.GetVFLinkName(netconf.DeviceID)` is called unconditionally here. `GetVFLinkName` reads `/sys/bus/pci/devices/<pci>/net/` to find a kernel network interface name. For VFs bound to the `vhost_vdpa` driver (rather than a kernel net driver), this sysfs directory may be empty, causing `GetVFLinkName` to fail and abort `CmdAdd`.

In contrast, `sriov.CmdAdd` (line 90) guards this call with `if !userspaceMode`. Should there be a similar guard here, or is it guaranteed that the VF's kernel net device is always present alongside the vdpa device for the supported hardware (ConnectX-6 Dx)?

phoracek: @bgartzi just for your consideration. I'm not sure this is correct

phoracek:
@bgartzi two small notes, the rest looks good to me, thanks!

pkg/veth/plugin.go
bgartzi: I believe having read somewhere that it was deferred just in case ipam failed. Of course refetchIface can fail too, my bad...

What I mean by this, is that rather than inline I'm moving these deferred calls right before the refetchIface call.

pkg/vdpa/plugin.go
bgartzi: Good catch. It's not currently a problem, but it might be if we consider supporting userspace vdpa/vduse. It seems I left it there just because.

bgartzi:
thanks a lot @phoracek. Updates addressing the comments just sent!

phoracek:
@SchSeba hey, this changes the SR-IOV part of the codebase. Let me know if you want to take a look. Otherwise I will be merging this next week.

phoracek:
Thank you again @bgartzi for this enhancement and for leaving the project in a better state!

PR #542, opened 2026-06-15
fix(plugin): honor OvnPort/MAC from CNI args.cni in StdinData

fivetime:

Fixes a silent integration bug between **ovs-cni** and **multus-cni** that prevents the `ovnPort` feature from working when invoked through the standard multus annotation path.

**Background.** ovs-cni's plugin reads `OvnPort` (and `MAC`) only from the `CNI_ARGS` environment variable via `cnitypes.LoadArgs(args.Args, &EnvArgs{})`. multus-cni, however, takes the `cni-args` map from a pod's `k8s.v1.cni.cncf.io/networks[].cni-args` annotation and injects it into the delegate plugin's **StdinData** under `args.cni.<Key>`, _not_ via `CNI_ARGS`. As a result:

- the veth is added to the bridge,
- `external_ids:iface-id` is **never set**,
- `ovn-controller` never binds the OVN logical switch port to the chassis,
- the pod has no L2/L3 connectivity through OVN.

There is no warning anywhere in the stack: multus passes its own `IgnoreUnknown=1` to delegates, so even if a user tries the lower-case `cni-args.ovnPort` key the unknown-args check is suppressed and the request appears to succeed.

**Fix.** Add an `Args` field to `NetConf` capturing the CNI 0.4.0+ `args.cni` map from StdinData, and have the veth/sriov/vdpa `CmdAdd` / `CmdDel` / `CmdCheck` paths fall back to it (via a new `common.ApplyConfArgsFallback` helper) when `CNI_ARGS` did not supply `OvnPort` or `MAC`. `CNI_ARGS` keeps priority, so callers that already pass values via env (e.g. kubevirt virt-launcher) are unaffected.

**This PR supersedes #529.** That PR was filed before the vhost_vdpa refactor (#427) split the plugin into per-backend packages; this rebased version applies the same fix in all three backends and addresses the automated review feedback on #529:

* `PluginArgs.Cni` is `map[string]json.RawMessage` instead of `map[string]string`, so an unrelated array/object entry under `args.cni` (for example the IPAM-style `ips` array) does not break `NetConf` unmarshal — values are decoded lazily only for the keys the fallback consumes.
* Key matching is **case-insensitive** — pod annotations written as `ovnPort`, `OvnPort`, or `ovnport` all resolve.
* The fallback helper accepts `nil` pointers for either output, so `CmdDel` / `CmdCheck` (which only care about `ovnPort`) need no dummy variables.

End-to-end verified on a real Neutron + OVN cluster (multus-cni 4.2.4, OVN 24.x, Open vSwitch 3.x):

1. Created Neutron ports on a **geneve tenant** network (`provider:network_type=geneve`) and a **flat provider** network (`provider:network_type=flat`, mapped to `br-ex` via OVN `bridge_mappings`).
2. Two pods attached to each network using the standard annotation:
   ```yaml
   k8s.v1.cni.cncf.io/networks: |
     [{"name":"ovn-tenant","mac":"fa:16:3e:...","ips":["192.168.1.8/24"],
       "cni-args":{"OvnPort":"<neutron-port-uuid>"}}]
   ```
3. **Before this patch** (v0.39.0): veths land on `br-int` with empty `external_ids`, OVN SB `port_binding.chassis=[]`, ping 100 % loss. The fix only worked when an operator manually ran `ovs-vsctl set interface <veth> external_ids:iface-id=<uuid>`.
4. **After this patch**: `external_ids:iface-id` and `ovn-installed=true` show up automatically, OVN SB binds both LSPs to the local chassis (`up=true`), and ping between pods works for both overlay (geneve) and underlay (flat localnet) — 0 % loss, RTT 0.1–7 ms.

The change is intentionally minimal: a `nil`-guarded helper plus nine one-line call sites (three backends x ADD/DEL/CHECK). CNI_ARGS still wins when both paths are present, so downstream consumers like kubevirt are not affected.

**Release note**:

```release-note
Fix `ovnPort` / `MAC` not being honored when passed via the multus-cni
`cni-args` pod annotation. ovs-cni now also reads these values from
`args.cni.<Key>` in the CNI conf StdinData, in addition to the existing
`CNI_ARGS` env var path. Matching is case-insensitive; CNI_ARGS retains
priority.
```

## Release Notes

* **New Features**
  * Added support for CNI 0.4.0+ per-invocation arguments, enabling automatic fallback for network parameters like `mac` and `ovnPort` when not provided via environment/config.

* **Improvements**
  * Improved consistency of `mac`/`ovnPort` initialization across veth, SR-IOV, and vDPA workflows, including DEL/CHECK paths.
  * Updated documentation to describe how to supply `mac` and `ovnPort` via `cni-args`/`CNI_ARGS`, including case-insensitive matching and handling of unrecognized/invalid entries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,10 @@ spec:
 EOF
 ```
 
+Per-pod parameters such as the interface `mac` or an OVN `ovnPort` can be passed
+through the network's `cni-args`; see
+[Per-Pod Arguments](docs/cni-plugin.md#per-pod-arguments-mac--ovnport) for details.
+
 Such pod should contain default `eth0` interface connected to default Kubernetes network and also `net1` connected to the bridge.
 
 ```shell
--- a/docs/cni-plugin.md
+++ b/docs/cni-plugin.md
@@ -62,6 +62,16 @@ Another example with a port which has an interface of type system:
 * `configuration_path` (optional): configuration file containing ovsdb
   socket file path, etc.
 
+The following are *per-invocation* arguments rather than static network
+configuration. They are not set in the `NetworkAttachmentDefinition` `config`
+body, but passed per pod through CNI args (see
+[Per-Pod Arguments](#per-pod-arguments-mac--ovnport)):
+
+* `mac` (string, optional): MAC address to assign to the container interface.
+* `ovnPort` (string, optional): value written to the OVS interface's
+  `external_ids:iface-id`, used to bind the port to an OVN logical switch port.
+  When set, `bridge` may be omitted (it is derived from the port).
+
 
 _*Note:* if `deviceID` is provided, then it is possible to omit `bridge` argument. Bridge will be automatically selected by the CNI plugin by following
 the chain: Virtual Function PCI address (provided in `deviceID` argument) > Physical Function > Bond interface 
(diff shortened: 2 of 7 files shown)


SchSeba (commented):
LGTM can you please extend the readme documentation to explain this new feature.

also another point that I am not sure here is if we want to have the `strings.ToLower` or be specific on the parameter name

fivetime:
@SchSeba thanks for the review!

I've pushed a follow-up commit (c2afce3) adding documentation for this feature — a new **Per-Pod Arguments (`mac` / `ovnPort`)** section in `docs/cni-plugin.md` (covering both the `CNI_ARGS` and the Multus `cni-args` → `args.cni` delivery paths, with a pod-annotation example), plus a pointer to it from the README.

**On `strings.ToLower` / case-insensitive matching** — I'd lean toward keeping it, for two reasons:

1. **It targets the exact failure mode this PR fixes.** The reason `OvnPort` didn't reach the plugin before was a *silent* mismatch — nothing errors out, the port just isn't wired. Case sensitivity is the same footgun one level up: a user writing `ovnPort` instead of `OvnPort` in their pod annotation gets no error and no effect, just a broken pod to debug. Being lenient here is consistent with the intent of the fix.
2. **This path is human-authored, unlike `CNI_ARGS`.** I agree the asymmetry with the `CNI_ARGS`/`EnvArgs` path is worth calling out — but I think it actually justifies the difference: `CNI_ARGS` keys are produced by the runtime from fixed struct field names, whereas `args.cni` keys come straight from a hand-written `k8s.v1.cni.cncf.io/networks[].cni-args` annotation. Leniency is most valuable precisely on the path people type by hand.

On determinism: the loop takes the first non-empty match and returns, and a single annotation realistically won't carry two differently-cased spellings of the same key — but if you'd prefer, I'm happy to add an explicit guard (or restrict to a small documented canonical set: `mac`, `ovnPort`) so behavior is well-defined even in that edge case.

If you'd still rather have strict exact-match, that's an easy change — just let me know and I'll switch it.

SchSeba (approved):
/lgtm
/approve

PR #564, opened 2026-09-09
Report configured MTU in CmdAdd result for all backends

omastama-spec:

ovs-cni configures interface MTU when set in the NAD but does not include it in the CNI result. This means Multus cannot advertise the MTU in the pod network-status annotation, and consumers of that annotation have no way to discover the configured MTU.

This PR sets hostIface.Mtu and contIface.Mtu in the CmdAdd return path for all backends (veth, vhost_vdpa, sriov) when MTU is configured and successfully applied.

Changes:
- veth: set hostIface.Mtu and contIface.Mtu after SetupVeth
- vdpa: set hostIface.Mtu and contIface.Mtu after MTU is applied to repLink and vfLink in setupKernelVdpaVhost
- sriov: set hostIface.Mtu in SetupSriovInterface after MTU is applied to the host representor, and contIface.Mtu in setupKernelSriovContIface after MTU is applied to the VF
- tests: add assertions to verify the CNI result includes MTU

- MTU is only advertised when mtu != 0 (user explicitly set it in the NAD)
- MTU is set after successful LinkSetMTU calls, so it is only reported if actually applied
- For sriov userspace mode, only hostIface.Mtu is set because there is no real container network interface
- Test assertions added to verify the result includes MTU for both host and container interfaces

**Release note**:
```release-note
Report configured MTU in CNI result for veth, vhost_vdpa, and sriov backends so Multus can advertise it in the network-status annotation
```

## Summary by CodeRabbit

* **Bug Fixes**
  * CNI results now correctly report the configured MTU for host and container interfaces across SR-IOV, vDPA, and veth setups.
  * MTU reporting is included when a non-zero MTU is configured, improving consistency between interface settings and reported results.

* **Tests**
  * Added validation to confirm reported MTU values for CNI versions 1.0.0 and newer.


bgartzi:
@omastama-spec this looks good: I tested it on a cluster and I could see the MTU reported on the network-status annotation of pods.

However, it seems there's still something off on the cni tests, they fail on the `Checking that host interface MTU in the result matches configured MTU` step.

omastama-spec:
> @omastama-spec this looks good: I tested it on a cluster and I could see the MTU reported on the network-status annotation of pods.
> 
> However, it seems there's still something off on the cni tests, they fail on the `Checking that host interface MTU in the result matches configured MTU` step.

Thanks for testing it on a real cluster @bgartzi ! The Mtu field only exists in CNI spec 1.0.0+, so the result loses it during version conversion for older specs (0.3.x, 0.4.x). i have pushed a fix that guards the MTU result assertions with a version check so they only run for CNI >= 1.0.0.

bgartzi:
Thanks @omastama-spec. However, c90a039e fixes something that ae3039a5 brings, so I think squashing them would be cleaner.

bgartzi:
/lgtm
@SchSeba, @phoracek could some of you follow with the ok-to-test and approve steps? Thanks!

omastama-spec:
The pull-e2e-ovs-cni failure looks unrelated to this PR  make lint fails because golangci-lint can't parse the Go stdlib (export data version 4 > maximum supported version 2), i.e. CI's Go is newer than the bundled golangci-lint supports. **It fails on pkg/cache/cache.go (untouched here)** so any PR hits it right now. Probably needs a Go pin or golangci-lint bump as a separate fix.

omastama-spec:
@phoracek gentle ping on this whenever you get a chance. I think it's blocked on /ok-to-test plus the unrelated lint/toolchain failure (CI's Go version vs the bundled golangci-lint). Happy to open a separate PR for the golangci-lint bump if that would help unblock CI. Thanks!

